PR #2, opened 2020-01-22
initial app and plugin APIs

albertshau:


delta-api/src/main/java/io/cdap/delta/api/EventConsumer.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package io.cdap.delta.api;
+
+/**
+ * Consumes events emitted by a source and applies them to the target system.
+ */
+public interface EventConsumer {
+
+  void start();
```
yaojiefeng, line 27: Need javadoc for this

yaojiefeng (approved):
I didn't go through the logic of the code, since the API and logic will change later. just one minor comment, LGTM

PR #9, opened 2020-01-27
Source table listing

albertshau:


delta-proto/src/main/java/io/cdap/delta/proto/CodedException.java
```diff
@@ -0,0 +1,38 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.delta.proto;
+
+/**
+ * An exception that contains an error code.
+ */
+public class CodedException extends RuntimeException {
```
yaojiefeng, line 22: Have a general question. Is this a better way to deal with the http code? Since I feel it is more straightforward to have each REST endpoint to explicitly send the error code. Now it adds a wrapper and we have to look at what `DeltaService` does to find out what exactly the exception it can throw.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -191,6 +195,6 @@ private void respond(String namespaceName, HttpServiceResponder responder, Names
     /**
      * Create the response that should be returned by the endpoint.
      */
-    void respond(Namespace namespace) throws Exception;
+    void respond(DraftService draftService, Namespace namespace) throws Exception;
```
yaojiefeng, line 198: Similar question here, is this interface a better way to implement the handler class?

delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
```diff
@@ -74,6 +77,70 @@ public boolean isService() {
...
+    return getStages().stream()
+      .filter(s -> DeltaTarget.PLUGIN_TYPE.equals(s.getPlugin().getType()))
+      .findFirst()
+      .orElseThrow(() -> new IllegalArgumentException("No source stage found."));
+  }
+
+  /**
+   * Validate that the stages contain all required fields and that there is a source defined.
+   *
+   * @return an array with the source stage as the first element and the target as the next element. The target
```
yaojiefeng: The javadoc seems accurate. The function just returns the targetStage

delta-proto/src/main/java/io/cdap/delta/proto/Plugin.java
```diff
@@ -44,11 +45,23 @@ public String getType() {
   }
 
   public Map<String, String> getProperties() {
-    return properties;
+    return properties == null ? Collections.emptyMap() : properties;
```
yaojiefeng, line 49: Can we have this logic in the constructor instead of get()

delta-storage/src/test/java/io/cdap/delta/store/DraftServiceTest.java
```diff
@@ -0,0 +1,196 @@
...
+  public void testListTablesFromNonexistantDraft() throws IOException {
+    DraftService service = new DraftService(getTransactionRunner());
+    service.listDraftTables(new DraftId(new Namespace("ns", 0L), "testListTablesFromNonexistantDraft"),
+                            new MockConfigurer(null, null));
+  }
+
+  /**
+   * Mock source that returns pre-determined table list and table detail.
+   */
+  private static class MockSource implements DeltaSource {
```
yaojiefeng: Do you think we can have this source as a separate class so other tests can use it?

yaojiefeng (commented):
Some questions

delta-proto/src/main/java/io/cdap/delta/proto/Plugin.java
albertshau: no because this is normally constructed through gson deserialization of user input. I can add comments so that's clear

delta-proto/src/main/java/io/cdap/delta/proto/CodedException.java
albertshau: I'm not sure what you mean by having the rest endpoint send the code. Do you mean not having the http code in the exception and have the endpoint catch the different exception types directly?

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
albertshau: Not quite sure what the question is, but what we're trying to avoid is having any application logic in the handler. The handler should only be responsible for decoding request bodies and things of that nature. This way the DraftService can be used in the handler, in a unit test, in some other service, or in some other handler, and we're ensured that it's the same logic.

delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
albertshau: oops, fixed

delta-storage/src/test/java/io/cdap/delta/store/DraftServiceTest.java
albertshau: moved MockSource and MockConfigurer to their own classes

yaojiefeng (approved):
lgtm

PR #11, opened 2020-01-29
implement assessment endpoints

albertshau:
Implement service endpoints that create a pipeline assessment and
table assessment for a draft.

Added a tables list to the DeltaConfig to configure which tables
the source should read from.


delta-api/src/main/java/io/cdap/delta/api/assessment/Support.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.delta.api.assessment;
+
+/**
+ * Support level of a feature.
```
yaojiefeng: I think more javadoc is needed here about what this enum is doing.

delta-storage/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -133,15 +151,142 @@ public TableList listDraftTables(DraftId draftId, Configurer configurer) throws
...
+    deltaConfig.validatePipeline();
+
+    TableRegistry tableRegistry = createTableRegistry(draft, configurer);
+    TableAssessor<List<ColumnDetail>> sourceTableAssessor = createTableAssessor(configurer, deltaConfig.getSource());
+    TableAssessor<Schema> targetTableAssessor = createTableAssessor(configurer, deltaConfig.getTarget());
+
+    List<Problem> connectivityIssues = new ArrayList<>();
+    List<TableSummaryAssessment> tableAssessments = new ArrayList<>();
+    // go through all tables that the pipeline should read, fetching detail about each of the tables
+    for (SourceTable sourceTable : draft.getConfig().getTables()) {
```
yaojiefeng: Does this mean the accessment can only be done if the draft provides the specifc table names? What will happen if the draft does not specify it and wants to read from all tables?

delta-api/src/main/java/io/cdap/delta/api/assessment/SourceTableDetail.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package io.cdap.delta.api.assessment;
+
+import java.util.List;
+
+/**
+ * Detailed information about a source table.
+ */
+public class SourceTableDetail extends TableDetail<List<ColumnDetail>> {
```
yaojiefeng: I don't get why we need a separate class for source, seems like this class is no different than `TableDetail`, we can just create`TableDetail` right?

delta-api/src/main/java/io/cdap/delta/api/assessment/StandardizedTableDetail.java
```diff
@@ -0,0 +1,31 @@
...
+package io.cdap.delta.api.assessment;
+
+import io.cdap.cdap.api.data.schema.Schema;
+
+import java.util.List;
+
+/**
+ * Standardized information about a source table.
+ */
+public class StandardizedTableDetail extends TableDetail<Schema> {
```
yaojiefeng: Same here

delta-storage/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -133,15 +151,142 @@ public TableList listDraftTables(DraftId draftId, Configurer configurer) throws
...
+  private TableAssessment merge(TableAssessment srcAssessment, TableAssessment targetAssessment) {
+    List<ColumnAssessment> columnAssessments = new ArrayList<>();
+    Iterator<ColumnAssessment> targetColumnAssessments = targetAssessment.getColumns().iterator();
+    for (ColumnAssessment columnAssessment : srcAssessment.getColumns()) {
+      columnAssessments.add(merge(columnAssessment, targetColumnAssessments.next()));
+    }
+    return new TableAssessment(columnAssessments);
+  }
+
+  private ColumnAssessment merge(ColumnAssessment srcAssessment, ColumnAssessment targetAssessment) {
```
yaojiefeng, line 298: More javadoc help is helpful to understand what the merge is doing

yaojiefeng (commented):
Some questions

delta-storage/src/main/java/io/cdap/delta/store/DraftService.java
albertshau: The current mocks force the user to select tables, so there isn't a planned way to tell it to read all tables in a database. 
That being said, it's pretty straightforward to change it such that specifying nothing means reading everything, so I'll modify that.

delta-api/src/main/java/io/cdap/delta/api/assessment/SourceTableDetail.java
albertshau: the difference is that the source one represents the columns differently, since we want to use the SQL names for the column types and not the CDAP Schema types. The two classes are confusing though, so I'll combine them into a single one that has both a `List<ColumnDetail>` and a `Schema`.

albertshau: actually, I can't combine them because there are cases when the `List<ColumnDetail>` contains types that are not supported by the source, usually because there isn't a good representation for them in CDAP Schema. Geo objects are an example of this.

I think the confusing part is that TableDetail is generic. I'm just going to remove that class

albertshau:
https://builds.cask.co/browse/DEL-BT6

yaojiefeng (approved):

PR #12, opened 2020-01-30
add unit test framework and corresponding worker tests

albertshau:
Added a delta-test module that includes a mock source and mock
target that can be used for testing the app and plugins. Added
a couple unit tests for the DeltaWorker using the test module.


albertshau:
https://builds.cask.co/browse/DEL-BT7

CuriousVini (approved):

PR #15, opened 2020-01-31
add macro support

albertshau:


albertshau:
https://builds.cask.co/browse/DEL-BT10

CuriousVini (approved):
lgtm, just one question - why `DefaultMacroEvaluator.java` has 2 copies? should we create a common module and rely on that in delta-app and delta-storage?

albertshau:
oops that was accidental after some branch rebasing, will remove

PR #17, opened 2020-01-31
moved classes around and removed delta-storage module

albertshau:
delta-storage was added because of hadoop version conflicts
between the app and CDAP unit-test module. Those have since been
resolved, so moved storage classes back into delta-app, and moved
a couple mock classes for testing into the new delta-test module.


CuriousVini (approved):

albertshau:
https://builds.cask.co/browse/DEL-BT12

PR #20, opened 2020-02-04
fix the table assessment endpoint response

albertshau:
Fixed the table assessment endpoint to return an object that
contains both the column name and type as seen by both the source
and target. Added a FullTableAssessment class to represent this
response. Also add a list of problems to the TableAssessment
class so that a target can record that it needs a primary key,
and other types of similar problems

### DIFF
--- a/delta-api/src/main/java/io/cdap/delta/api/assessment/ColumnAssessment.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/assessment/ColumnAssessment.java
@@ -25,35 +25,52 @@ import javax.annotation.Nullable;
 public class ColumnAssessment {
   private final String name;
   private final String type;
+  private final String sourceName;
   private final ColumnSupport support;
   private final ColumnSuggestion suggestion;
 
-  public ColumnAssessment(String name, String type) {
+  private ColumnAssessment(ColumnSupport support, String name, String type,
+                           @Nullable String sourceName, @Nullable ColumnSuggestion suggestion) {
     this.name = name;
     this.type = type;
-    this.support = ColumnSupport.YES;
-    this.suggestion = null;
-  }
-
-  public ColumnAssessment(String name, String type, ColumnSupport support, ColumnSuggestion suggestion) {
-    this.name = name;
-    this.type = type;
+    this.sourceName = sourceName;
     this.support = support;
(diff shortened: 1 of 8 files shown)


delta-api/src/main/java/io/cdap/delta/api/assessment/TableAssessment.java
```diff
@@ -26,15 +26,21 @@
  */
 public class TableAssessment {
   private final List<ColumnAssessment> columns;
+  private final List<Problem> features;
```
yaojiefeng, line 29: Would it be better to call this `problems` to be more intuitive?

delta-proto/src/main/java/io/cdap/delta/proto/FullTableAssessment.java
```diff
@@ -0,0 +1,61 @@
...
+
+import io.cdap.delta.api.assessment.Problem;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Full table assessment using assessments from both the source and target.
+ */
+public class FullTableAssessment {
```
yaojiefeng: I feel `FullAssessment` is a better name for this since this contains information about both tables. On a first glance, I think this class is still about one table, but provide more information than `TableAssessment`

yaojiefeng (approved):
minor comments

delta-api/src/main/java/io/cdap/delta/api/assessment/TableAssessment.java
albertshau: the mocks have a 'features' list of problems and a 'connectivity' list of problems, which is where this name comes from, though I agree it's confusing. I'll rename to featureProblems.

delta-proto/src/main/java/io/cdap/delta/proto/FullTableAssessment.java
albertshau: It's still about one table, though it can contain transformed columns due to schema normalization (and in the future, actual user specified transforms). FullAssessment makes me think of an assessment for the entire pipeline. I think I'll rename this to TableAssessmentResponse, since it's only used as the response for the table assess endpoint.

PR #21, opened 2020-02-05
CDAP-16244 add metrics for events applied

albertshau:
Added methods to the DeltaTargetContext to increment in-memory
counts for the number of DML and DDL operations replicated.
These counts are emitted as metrics after an offset is committed.
    
Target plugins must correctly increment these counters in order
for the metrics to be accurate. This cannot be done by the app
because the app does not have control over when the plugin
decides to commit the offset.

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/app/EventMetrics.java
+++ b/delta-app/src/main/java/io/cdap/delta/app/EventMetrics.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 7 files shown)


albertshau:
please don't review yet, this approach doesn't actually work all the time and requires changes.

albertshau:
https://builds.cask.co/browse/DEL-BT16

delta-app/src/main/java/io/cdap/delta/app/EventMetrics.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public void incrementDDLCount() {
+    ddlEventCount++;
+  }
+
+  public void emitMetrics() {
+    for (DMLOperation op : dmlEventCounts.keySet()) {
+      metrics.count(String.format("%s.dml.%s", prefix, op.name().toLowerCase()), dmlEventCounts.get(op));
+      dmlEventCounts.put(op, 0);
```
CuriousVini, line 53: Is it possible that caller keeps calling incrementCount while emitMetrics is being reset? Does this class need to be thread safe?

CuriousVini (approved):
one question, code changes lgtm

delta-app/src/main/java/io/cdap/delta/app/EventMetrics.java
albertshau: good catch, it will cause weird things to happen. I'll just make these methods synchronized.

PR #22, opened 2020-02-05
CDAP-16246 add event type filtering

albertshau:
Enhanced the config to allow the user to specify a blacklist for
DML and DDL event types. A blacklist can be set at the pipeline
level for all tables, or for a specific table.

Refactored the source API to take a new EventReaderDefinition
class that contains information about what to read. This is more
extensible, as it allows us to add more constraints without breaking
backwards compatibility in the API.

### DIFF
--- a/delta-api/src/main/java/io/cdap/delta/api/SourceTable.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/SourceTable.java
@@ -17,8 +17,10 @@
 package io.cdap.delta.api;
 
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 /**
  * Object representing what data from a table should be read by the source.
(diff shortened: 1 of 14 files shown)


delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -49,7 +47,7 @@
   private static final String TARGET = "target";
   private static final String GENERATION = "generation";
   private static final String OFFSET_BASE_PATH = "offset.base.path";
-  private static final String SOURCE_TABLES = "source.tables";
+  private static final String READER_DEFINITION = "source.tables";
```
shifuxu0301: ditto, should we change the value also to "reader.definition" for consistency?

delta-api/src/main/java/io/cdap/delta/api/SourceTable.java
```diff
@@ -46,8 +57,26 @@ public String getTable() {
...
   }
 
-  public List<SourceColumn> getColumns() {
-    return columns;
+  /**
+   * @return table columns that should be replicated. If empty, all columns should be replicated.
+   */
+  public Set<SourceColumn> getColumns() {
+    // check for null because this object can be created through deserialization of user provided input
+    return columns == null ? Collections.emptySet() : columns;
```
shifuxu0301: Should it be good to make it safe to return unmodified empty set if columns is null?

delta-api/src/main/java/io/cdap/delta/api/SourceTable.java
```diff
@@ -46,8 +57,26 @@ public String getTable() {
...
+  }
+
+  /**
+   * @return set of DML operations that should be ignored
+   */
+  public Set<DMLOperation> getDmlBlacklist() {
+    return dmlBlacklist == null ? Collections.emptySet() : dmlBlacklist;
```
shifuxu0301: same thing above.

delta-api/src/main/java/io/cdap/delta/api/SourceTable.java
```diff
@@ -46,8 +57,26 @@ public String getTable() {
...
+  }
+
+  /**
+   * @return set of DDL operations that should be ignored
+   */
+  public Set<DDLOperation> getDdlBlacklist() {
+    return ddlBlacklist == null ? Collections.emptySet() : ddlBlacklist;
```
shifuxu0301: same thing here.

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
albertshau: oops, yes will do that

delta-api/src/main/java/io/cdap/delta/api/SourceTable.java
albertshau: Collections.emptySet() is unmodifiable, but you're right, it's better to wrap the columns variable to make it unmodifiable.

Was wrapping it in the constructor, but the constructor is not called when this is created through deserialization. Will adjust several of these classes accordingly.

shifuxu0301 (approved):

PR #23, opened 2020-02-06
fix unit test

albertshau:
Also allow APIs to throw exception, since we found the plugin implementations are often wrapping their checked exceptions in runtime exceptions.


albertshau:
https://builds.cask.co/browse/DEL-BT18

yaojiefeng (approved):

PR #24, opened 2020-02-06
add schema field to table detail and source table

yaojiefeng:
Add a schema field to table detail and source table. This is required since for SQLServer, the full table name has to be passed in to Debizium as [schemaName].[tableName] to be able to read records from it. I think it is similar case for Oracle. We cannot simply return the full name in `TableSummary` because when jdbc tries to fetch the metadata, using full table name will not return anything. So adding this to SourceTable so that we can get it back to construct the full table name.

### DIFF
--- a/delta-api/src/main/java/io/cdap/delta/api/SourceTable.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/SourceTable.java
@@ -21,6 +21,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import javax.annotation.Nullable;
 
 /**
  * Object representing what data from a table should be read by the source.
@@ -28,6 +29,10 @@ import java.util.Set;
 public class SourceTable {
   private final String database;
   private final String table;
+  // this schema is required for some db to uniquely identify the table, if this is not provided, it will not
+  // be able to fetch the records
+  // TODO: CDAP-16261 have a better way to identify the table
+  private final String schema;
   private final Set<SourceColumn> columns;
   private final Set<DMLOperation> dmlBlacklist;
   private final Set<DDLOperation> ddlBlacklist;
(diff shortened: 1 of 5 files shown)


delta-api/src/main/java/io/cdap/delta/api/SourceTable.java
```diff
@@ -21,13 +21,17 @@
...
 
 /**
  * Object representing what data from a table should be read by the source.
  */
 public class SourceTable {
   private final String database;
   private final String table;
+  // this schema is required for some db to uniquely identify the table, if this is not provided, it will not
+  // be able to fetch the records
+  private final String schema;
```
albertshau, line 35: This will also need to be added to 'DBTable', the change event classes, and anywhere else where a table is identified.

we probably need to rethink how a source 'table' is identified as we add more sources.

albertshau:
from the other PR, it seems like it's ok if the table name is [schema].[table]? Is this still needed?

I think we'll probably need to change how a 'table' is identified beyond just introducing a 'schema' field, but wondering if we can do that later after we think about all the planned sources (unless it is required now).

yaojiefeng:
Based on https://github.com/data-integrations/database-delta-plugins/pull/9#discussion_r376060433, the table name cannot be `[schema].[table]`, that's why I add this field since it is currently required to make the registry work

delta-api/src/main/java/io/cdap/delta/api/SourceTable.java
yaojiefeng: I was thinking to do that last night but didn't do that. Since this schema field is only needed in the `SourceTable` when passing the full table name in Debizium config. Adding this to all table related classes means we also need to add this to `DDLEvent`, etc. And I don't think this is a good way to handle this kind of problem so this pr is like a workaround, maybe I will open a JIRA to track this and fix it in the future

yaojiefeng: Created, https://issues.cask.co/browse/CDAP-16261

albertshau (approved):

PR #26, opened 2020-02-07
CDAP-16245 add table state apis

albertshau:
Added methods to the source and target contexts to allow them to
record state information. The source can record that the pipeline
is snapshotting, replicating, or failing. The target can record
that specific tables are ok or failing. If something is failing,
a replication error containing a message and a stack trace can
also be provided.

Also added a REST endpoint to get the state for a pipeline.

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/app/DeltaPipelineId.java
+++ b/delta-app/src/main/java/io/cdap/delta/app/DeltaPipelineId.java
@@ -17,14 +17,15 @@
 package io.cdap.delta.app;
 
 /**
- * Uniquely identifies a delta pipeline
+ * Uniquely identifies a delta pipeline.
+ * The generation is used to distinguish between pipeline X that was created, then deleted, then created again.
  */
 public class DeltaPipelineId {
   private final String namespace;
   private final String app;
-  private final String generation;
+  private final long generation;
 
-  public DeltaPipelineId(String namespace, String app, String generation) {
+  public DeltaPipelineId(String namespace, String app, long generation) {
     this.namespace = namespace;
     this.app = app;
     this.generation = generation;
@@ -38,7 +39,7 @@ public class DeltaPipelineId {
     return app;
   }
 
-  public String getGeneration() {
+  public long getGeneration() {
     return generation;
   }
 }
--- a/delta-app/src/main/java/io/cdap/delta/app/OffsetAndSequence.java
+++ b/delta-app/src/main/java/io/cdap/delta/app/OffsetAndSequence.java
@@ -14,32 +14,27 @@
  * the License.
  */
 
-package io.cdap.delta.api;
+package io.cdap.delta.app;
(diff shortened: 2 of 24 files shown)


albertshau:
https://builds.cask.co/browse/DEL-BT21

delta-app/src/main/java/io/cdap/delta/app/DeltaPipelineId.java
```diff
@@ -22,9 +22,9 @@
 public class DeltaPipelineId {
   private final String namespace;
   private final String app;
-  private final String generation;
+  private final long generation;
 
-  public DeltaPipelineId(String namespace, String app, String generation) {
+  public DeltaPipelineId(String namespace, String app, long generation) {
```
shifuxu0301, line 28: One question, why we change the data type to long here?

delta-app/src/main/java/io/cdap/delta/app/OffsetAndSequence.java
```diff
@@ -0,0 +1,40 @@
...
+package io.cdap.delta.app;
+
+import io.cdap.delta.api.Offset;
+
+/**
+ * An offset and sequence number.
+ */
+public class OffsetAndSequence {
+  private final Offset offset;
+  private final long sequenceNumber;
```
shifuxu0301, line 26: One question here, why is the sequnceNumber for? Why we need to combine it with the offset?

shifuxu0301 (commented):
Overall LGTM, few questions attached on the comments.

delta-app/src/main/java/io/cdap/delta/app/DeltaPipelineId.java
albertshau: I'll add some comments on how it's being used, but basically we don't have a hook in CDAP that lets an application run some logic when it is deleted, which means we don't have a way to delete the state (offsets, etc). If a replicator is deleted, then re-created with the same name, we don't want to start from the old offset, so we generate this number and use it in the file path for the state. 

In order to figure out which is the current state, the generation needs to be monotonically increasing instead of a random string, which is why it's being changed here.

delta-app/src/main/java/io/cdap/delta/app/OffsetAndSequence.java
albertshau: I'll add some comments here too.

sequence number is a piece of information used by the target to ensure changes are applied correctly. Basically, if data with a higher sequence number already exists in the table, we assume the change was already applied in the past and can be ignored. You can see an example of how it is used at https://wiki.cask.co/display/CE/Change+Data+Capture#ChangeDataCapture-Example.2.  It needs to be stored with the offset because the event for that offset must match the same sequence number, otherwise the target logic no longer makes sense.

shifuxu0301 (approved):

PR #32, opened 2020-02-12
CDAP-16251 retry failures up to a limit

albertshau:
Refactored the worker so that events emitted by the source are
first added to a work queue instead of directly passed to the
target. This is required so that the app is in control of when
calls to apply a DMLEvent or DDLEvent are made, and can properly
handle any exceptions that are thrown if they cannot be applied.

If an exception is thrown when a DML or DDL event is being applied,
the worker will stop the event reader and consumer, reset the offset
and sequence number to the last committed offset, then start the
reader from that old offset. The worker will continue to retry
until a configurable time limit is hit.

Also added some more localized retries around storing offsets
and pipeline state.


albertshau:
https://builds.cask.co/browse/DEL-BT25

delta-app/src/main/java/io/cdap/delta/app/DeltaContext.java
```diff
@@ -39,23 +44,34 @@
...
     this.id = id;
     this.runId = runId;
     this.metrics = metrics;
     this.stateStore = stateStore;
     this.pluginContext = pluginContext;
     this.eventMetrics = eventMetrics;
     this.stateService = stateService;
+    this.commitOffsetRetryPolicy = new RetryPolicy<>()
+      .withBackoff(1, 120, ChronoUnit.SECONDS)
+      .withMaxAttempts(Integer.MAX_VALUE)
```
shifuxu0301: Should we have an upper bound for the max num of backoff retry? The current max num of attempts are so high, what if something bad happened, the app will stuck for very long time. Looks like it is better to put some reasonable upper bound, so we can fail the pipeline quicker.

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -17,133 +17,284 @@
...
+    // of the loop. However, it is also possible for the worker to be stopped before it gets here, in which case it
+    // should just end immediately.
+    while (!shouldStop) {
+
+      // create the retry policy for handling exceptions thrown from calls to applyEvent()
+      // the worker will retry up to the configured maximum amount of time.
+      // On each failed attempt, offset will be rolled back to the last commit, and any in-memory state
+      // related to metrics and queued change events will be cleared away.
+      RetryPolicy<Object> retryPolicy = new RetryPolicy<>()
+        .withMaxAttempts(Integer.MAX_VALUE)
```
shifuxu0301, line 173: Seems we are going to be capped by max duration, do we have to provide max attempts?

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -17,133 +17,284 @@
...
+        .withMaxDuration(Duration.of(maxRetrySeconds, ChronoUnit.SECONDS))
+        .withDelay(retryDelaySeconds == 0 ? Duration.of(1, ChronoUnit.MILLIS) :
```
shifuxu0301, line 175: So this is not backoff retry, right?

delta-app/src/main/java/io/cdap/delta/app/QueueingEventEmitter.java
```diff
@@ -75,12 +72,10 @@ public void emit(DMLEvent event) {
...
 
-    sequenceNumber++;
     try {
-      consumer.applyDML(new Sequenced<>(event, sequenceNumber));
-    } catch (Exception e) {
-      // TODO: (CDAP-16251) retry
-      throw new RuntimeException(e);
+      eventQueue.put(new Sequenced<>(event, ++sequenceNumber));
+    } catch (InterruptedException e) {
+      // this means the worker is being stopped, so drop the change and let things stop
```
shifuxu0301, line 78: Should we at least log a warning for this exception?

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
albertshau: i'll add a comment, but if you don't set this, it defaults to 3 max attempts

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
albertshau: it's not because it's simpler for a user to specify a single delay. We can extend it in the future to different types if we need to.

delta-app/src/main/java/io/cdap/delta/app/QueueingEventEmitter.java
albertshau: The contract in our plugins is that events are emitted at least once, so this is a normal and acceptable thing to happen since it will be re-read when the pipeline is started again.

A warning usually indicates that somebody might have to do something to address the problem, which isn't the case here.

delta-app/src/main/java/io/cdap/delta/app/DeltaContext.java
albertshau: yeah I changed my mind a few times about what the right thing to do here would be. The problem is that there isn't a good way to fail the pipeline when this fails, because commitOffset() is called by the target plugin. So the app doesn't have any way to fail the pipeline here.

Now I'm thinking that it might be better to remove the retry and let the plugin handle it however it wants.

albertshau: Going to remove the retry here and force the plugin to handle it. The BQ target isn't going to want to throw the IOException from applyDDL or applyDML, as that would cause offset to be reset, which would cause a batch of changes to be replayed before failing again on commit and repeating the process forever.

To handle this, I'm going to add a special type of Exception that plugins can throw that will tell the app not to retry and fail immediately. This will allow the target plugin to retry the commit themselves for some time before throwing that special exception and failing the pipeline.

shifuxu0301: Sounds good.

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
shifuxu0301: Sure.

delta-app/src/main/java/io/cdap/delta/app/QueueingEventEmitter.java
shifuxu0301: makes sense.

shifuxu0301 (approved):

PR #33, opened 2020-02-13
CDAP-16262 Add precision and scale field into ColumnDetail.

shifuxu0301:
This pr is for adding fields in Column detail to support precision and scale for plugins.

### DIFF
--- a/delta-api/src/main/java/io/cdap/delta/api/assessment/ColumnDetail.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/assessment/ColumnDetail.java
@@ -17,9 +17,9 @@
 package io.cdap.delta.api.assessment;
 
 import java.sql.SQLType;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
-import javax.annotation.Nullable;
 
 /**
  * Details about a table column.
@@ -30,7 +30,14 @@ public class ColumnDetail {
   private final boolean nullable;
   private final Map<String, String> properties;
 
-  public ColumnDetail(String name, SQLType type, boolean nullable, @Nullable Map<String, String> properties) {
+  public ColumnDetail(String name, SQLType type, boolean nullable) {
+    this.name = name;
+    this.type = type;
+    this.nullable = nullable;
+    this.properties = new HashMap<>();
+  }
+
+  public ColumnDetail(String name, SQLType type, boolean nullable, Map<String, String> properties) {
     this.name = name;
     this.type = type;
     this.nullable = nullable;
@@ -49,7 +56,6 @@ public class ColumnDetail {
     return nullable;
   }
 
-  @Nullable
   public Map<String, String> getProperties() {
     return properties;
   }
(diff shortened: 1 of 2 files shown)


delta-api/src/main/java/io/cdap/delta/api/assessment/ColumnDetail.java
```diff
@@ -26,11 +26,15 @@
   private final String name;
   private final SQLType type;
   private final boolean nullable;
+  private final int length;
+  private final int scale;
 
-  public ColumnDetail(String name, SQLType type, boolean nullable) {
+  public ColumnDetail(String name, SQLType type, boolean nullable, int length, int scale) {
```
albertshau: are these things not present within the type?

This also won't be enough for other more complex types, such as geo types. It may be better to just have a Map<String, String> properties here, to let the TableRegistry set whatever it wants

shifuxu0301: If you mean SQLType? Yes. these are not present with that, these information will only be bind with the metadata.

Adding a properties map is good call. It will be more flexible. Will do that.

delta-api/src/main/java/io/cdap/delta/api/assessment/ColumnDetail.java
```diff
@@ -26,15 +28,13 @@
   private final String name;
   private final SQLType type;
   private final boolean nullable;
-  private final int length;
-  private final int scale;
+  private final Map<String, String> properties;
 
-  public ColumnDetail(String name, SQLType type, boolean nullable, int length, int scale) {
+  public ColumnDetail(String name, SQLType type, boolean nullable, @Nullable Map<String, String> properties) {
```
albertshau: In general it's better not to allow null collections and use empty collections instead.

Also, let's have 2 constructors, one that takes a Map and one that doesn't. Most types won't use properties, and this also won't break compilation of plugins.

delta-api/src/main/java/io/cdap/delta/api/assessment/ColumnDetail.java
```diff
@@ -26,11 +28,20 @@
   private final String name;
   private final SQLType type;
   private final boolean nullable;
+  private final Map<String, String> properties;
 
   public ColumnDetail(String name, SQLType type, boolean nullable) {
     this.name = name;
```
albertshau: this constructor should be changed to call the other one.

delta-api/src/main/java/io/cdap/delta/api/assessment/ColumnDetail.java
```diff
@@ -26,11 +28,20 @@
...
     this.type = type;
     this.nullable = nullable;
+    this.properties = new HashMap<>();
+  }
+
+  public ColumnDetail(String name, SQLType type, boolean nullable, Map<String, String> properties) {
+    this.name = name;
+    this.type = type;
+    this.nullable = nullable;
+    this.properties = properties;
```
albertshau: To protect against bad plugin behavior, should make a copy of the input map and wrap it in an unmodifiable one.

albertshau (approved):
couple small comments, otherwise lgtm.

PR #36, opened 2020-02-20
CDAP-16313 add parallelism

albertshau:
Added a parallelism config to the delta config that lets the user
specify either the number of worker instances to use. At deployment
time, the source tables are assigned to different worker instances
in a round robin fashion. Alternatively, the user can also manually
provide the table assignments instead of just the number of worker
instances.

Refactored the worker such that each worker instance only reads and
applies changes to its assigned set of tables. Offsets, sequence
numbers, and state were all changed to be per worker instance.

Currently, there is no way to change the number of instances
after the app is deployed, due to the tables being assigned at
deployment time and because there isn't currently a way to merge
different instance offsets/states.

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
+++ b/delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
@@ -16,7 +16,9 @@
 
 package io.cdap.delta.app;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
 import io.cdap.cdap.api.app.ApplicationSpecification;
 import io.cdap.cdap.api.macro.MacroEvaluator;
 import io.cdap.cdap.api.metrics.Metrics;
@@ -37,6 +39,9 @@ import io.cdap.delta.api.Offset;
 import io.cdap.delta.api.Sequenced;
 import io.cdap.delta.api.SourceTable;
 import io.cdap.delta.proto.DeltaConfig;
+import io.cdap.delta.proto.InstanceConfig;
+import io.cdap.delta.proto.ParallelismConfig;
+import io.cdap.delta.proto.TableId;
 import io.cdap.delta.store.DefaultMacroEvaluator;
 import io.cdap.delta.store.StateStore;
 import net.jodah.failsafe.Failsafe;
@@ -47,10 +52,12 @@ import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.lang.reflect.Type;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ArrayBlockingQueue;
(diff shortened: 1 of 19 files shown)


delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -139,17 +176,30 @@ public void initialize(WorkerContext context) throws Exception {
...
 
-    // TODO: make the queue size configurable? record metrics about queue size?
     eventQueue = new ArrayBlockingQueue<>(100);
   }
 
   @Override
   public void run() {
+    // handle the situation where the user manually increased the number of worker instances after
+    // the pipeline was deployed.
+    int instanceId = getContext().getInstanceId();
```
yaojiefeng, line 199: I am not quite familiar with the worker logic, can we just stop this instance if there is nothing to process?

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -103,14 +130,23 @@ public void initialize(WorkerContext context) throws Exception {
     long generation = Long.parseLong(context.getSpecification().getProperty(GENERATION));
 
     ApplicationSpecification appSpec = context.getApplicationSpecification();
-    DeltaConfig config = GSON.fromJson(appSpec.getConfiguration(), DeltaConfig.class);
+    config = GSON.fromJson(appSpec.getConfiguration(), DeltaConfig.class);
+
+    int instanceId = context.getInstanceId();
+    tableAssignments = GSON.fromJson(context.getSpecification().getProperty(TABLE_ASSIGNMENTS),
+                                     new TypeToken<Map<Integer, Set<TableId>>>() { }.getType());
```
yaojiefeng: This can be a static field

delta-app/src/main/java/io/cdap/delta/store/StateStore.java
```diff
@@ -95,8 +100,31 @@ public Long getLatestGeneration(String namespace, String pipelineName) throws IO
     return maxGeneration > 0 ? maxGeneration : null;
   }
 
+  public Collection<Integer> getWorkerInstances(DeltaPipelineId pipelineId) throws IOException {
```
yaojiefeng, line 103: I might miss the write path. Do we have the logic of writing the worker instances to the file?

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -316,4 +367,32 @@ private void startFromLastCommit() throws Exception {
...
+    ParallelismConfig parallelism = config.getParallelism();
+    List<InstanceConfig> instances = parallelism.getInstances();
+    Map<Integer, Set<TableId>> assignments = new HashMap<>();
+    if (!instances.isEmpty()) {
+      int instanceNum = 0;
+      for (InstanceConfig instanceConfig : instances) {
+        assignments.put(instanceNum, instanceConfig.getTables());
+        instanceNum++;
+      }
+      return assignments;
```
yaojiefeng, line 385: So here if `numInstances` > `instances.size()`, the instance number will be `instances.size()`. Should we log something since in UI user will see the number of instances is smaller than they specified.

yaojiefeng (commented):
some small comments

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
albertshau: yeah should probably just let it end, will change this.

delta-app/src/main/java/io/cdap/delta/store/StateStore.java
albertshau: yeah it's from changing the other methods in this class to take a DeltaWorkerId, and modify getPath() to include the instance id in the path.

yaojiefeng (approved):

albertshau:
https://builds.cask.co/browse/DEL-BT31

PR #37, opened 2020-02-27
CDAP-16343 fix filesystem usage

albertshau:
Create the FileSystem object using the URI from the path instead
of getting the default FileSystem. This properly allows it to
use a distributed filesystem instead of the local one.


yaojiefeng (approved):

albertshau:
https://builds.cask.co/browse/DEL-BT32

PR #38, opened 2020-02-27
Fix the state endpoint

albertshau:
Fixed the state endpoint to properly be annotated with the request
type. Also changed the design a little bit to be a POST with a
request body, to allow the caller to pass in an offset base path.
This is required because the base path is set per replicator.


elfenheart (approved):
👍 LGTM

PR #39, opened 2020-02-28
CDAP-16346 added a way for sources to notify failures

albertshau:
Added a notifyFailure method to the DeltaSourceContext that allows
an EventReader to notify the app that it is failing and tell the
app to reset state to the last committed offset.

### DIFF
--- a/delta-test/src/main/java/io/cdap/delta/test/mock/FailureEventReader.java
+++ b/delta-test/src/main/java/io/cdap/delta/test/mock/FailureEventReader.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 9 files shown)


shifuxu0301 (approved):

delta-test/src/main/java/io/cdap/delta/test/mock/FailureEventReader.java
```diff
@@ -0,0 +1,52 @@
...
+    try {
+      context.setError(new ReplicationError(exception));
+    } catch (IOException e) {
+      // shouldn't happen in unit tests
+    }
+    context.notifyFailed(exception);
+  }
+
+  @Override
+  public void stop() {
```
CuriousVini: why not add default no op implementation in in the interface?

CuriousVini (approved):
one comment. lgtm

albertshau:
https://builds.cask.co/browse/DEL-BT34

PR #40, opened 2020-03-05
change offset value type to string

albertshau:
changed the offset values to be strings to make it more user
friendly to modify the offset in the future

### DIFF
--- a/delta-api/src/main/java/io/cdap/delta/api/Offset.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/Offset.java
@@ -16,7 +16,6 @@
 
 package io.cdap.delta.api;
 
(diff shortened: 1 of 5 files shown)


shifuxu0301 (approved):
LGTM for this PR. 
One question may not directly related to this PR. So far, it seems like we do not have any data corruption protection when we wrote offset into disk, do we plan to add maybe MD5 in near future in the app?

albertshau:
we could add some, though what would we be protecting against? Somebody manually modifying the files?

shifuxu0301:
One case I had met before is the bit flip on disk causing the file corrupted, the possibility is pretty low though. And checksum may not fully mitigate this issue.

PR #43, opened 2020-03-17
CDAP-16347 Added list of missing features for TableDetail in delta api.

shifuxu0301:
This PR is for adding list of missing features for TableDetail class, which will be used to represent the case that CDC is not enabled for table during the assessment. Source plugin will be the one to add missing feature problems when it detects cdc was not enabled during assessment table.


delta-api/src/main/java/io/cdap/delta/api/assessment/TableRegistry.java
```diff
@@ -37,9 +37,11 @@
    * @param table the table name
    * @return detail about the table
    * @throws TableNotFoundException if the specified table does not exist
+   * @throws TableCDCNotEnabledException if the CDC feature for the specified table is not enabled
    * @throws IOException if the table information could not be read
    */
-  TableDetail describeTable(String database, String table) throws TableNotFoundException, IOException;
+  TableDetail describeTable(String database, String table)
+    throws TableNotFoundException, TableCDCNotEnabledException, IOException;
```
albertshau: I don't think this is a good pattern, we'll end up throwing more and more different types of exceptions. Also, it means the information about the table cannot be returned, even though in cases like this, they actually can.

I think it's better to add a list of features and a list of connectivity problems to TableDetail, that then get added to the overall pipeline's problems. I would consider this a 'feature' problem, not a connectivity one.

shifuxu0301: Yes. I agreed this is not a good pattern for adding new problems in the future. So, currently the 'feature' and 'connectivity' problems are came together with PipelineAssessment, we can definitely do some refactor by binding problems into TableDetail and then merged all of them to PipelineAssessment. I will do that.

shifuxu0301: In order the keep the PR simple, I added the List of missing features with TableDetail class to accomplish the goal of checking if cdc is enabled during table assessment. There is additional discussion undergoing about should we also add list of connectivity problem into TableDetail class, will work on that in coming PRs if we decided to do so. So far, this PR will only focus on adding missing features so far.

albertshau (approved):

PR #45, opened 2020-03-24
CDAP-16475 Filter out unselected primary keys for table detail before passing to table assessment

shifuxu0301:
The delta app should also favor selected/unselected primary keys and use that to from filtered table detail before passing it for table assessment.

Tested on local, missing primary key showed up correctly if I manually unselected primary key during table listing.

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/store/DraftService.java
+++ b/delta-app/src/main/java/io/cdap/delta/store/DraftService.java
@@ -291,17 +291,20 @@ public class DraftService {
     // fetch detail about the table, then filter out columns that will not be read by the source
     TableDetail detail = tableRegistry.describeTable(db, table);
     List<Problem> missingFeatures = new ArrayList<>(detail.getFeatures());
-    detail.getPrimaryKey().stream()
+    List<String> unselectedPrimaryKeys = detail.getPrimaryKey().stream()
       // if there are no columns specified, it means this primary key will be read/selected
       .filter(columnWhitelist.isEmpty() ? col -> false : col -> !columnWhitelist.contains(col))
-      // add feature problem for any unselected primary key for table
-      .forEach(col -> missingFeatures.add(
+      .collect(Collectors.toList());
+    if (!unselectedPrimaryKeys.isEmpty()) {
+      missingFeatures.add(
         new Problem("Missing Primary Key",
-                    String.format("Column '%s' is part of the primary key for table '%s' in database '%s', but is " +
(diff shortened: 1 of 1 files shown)


delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -290,12 +290,16 @@ private TableAssessmentResponse assessTable(SourceTable sourceTable, TableRegist
 
     // fetch detail about the table, then filter out columns that will not be read by the source
     TableDetail detail = tableRegistry.describeTable(db, table);
+    List<String> selectedPrimaryKeys = detail.getPrimaryKey().stream()
```
albertshau: This should show up as a feature problem in the assessment. Something like:

'Column 'name' is part of the primary key for table 'db.table', but is not one of the selected fields to be replicated. This can result in different data at the target than at the source.

shifuxu0301: True. Totally agree. we should distinguish between the real root cause. Will do.

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -290,18 +290,26 @@ private TableAssessmentResponse assessTable(SourceTable sourceTable, TableRegist
...
-    List<String> selectedPrimaryKeys = detail.getPrimaryKey().stream()
-      // if there are no columns specified, it means all columns (including primary keys) should be read
-      .filter(columnWhitelist.isEmpty() ? col -> true : columnWhitelist::contains)
-      .collect(Collectors.toList());
+    List<Problem> missingFeatures = new ArrayList<>(detail.getFeatures());
+    detail.getPrimaryKey().stream()
+      // if there are no columns specified, it means this primary key will be read/selected
+      .filter(columnWhitelist.isEmpty() ? col -> false : col -> !columnWhitelist.contains(col))
+      // add feature problem for any unselected primary key for table
+      .forEach(col -> missingFeatures.add(
```
albertshau: nit: I think it's better to have a single problem that mentions all of the columns, otherwise it could look kind of silly if it's missing 8 columns or something like that.

shifuxu0301: Sure, since we do have primary key tag during table listing page. I agree to add just one problem here. I will do that change.

shifuxu0301:
Comment addressed, could u please take a look again.

PR #47, opened 2020-03-26
CDAP-16475 Add feature problem for unselected primary key during assessment.

shifuxu0301:
The delta app should also check unselected primary keys, if there are, then we should add it as missing feature problem. 

Tested on local, missing primary key showed up correctly if I manually unselected primary key during table listing.

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/store/DraftService.java
+++ b/delta-app/src/main/java/io/cdap/delta/store/DraftService.java
@@ -290,6 +290,21 @@ public class DraftService {
 
     // fetch detail about the table, then filter out columns that will not be read by the source
     TableDetail detail = tableRegistry.describeTable(db, table);
+    List<Problem> missingFeatures = new ArrayList<>(detail.getFeatures());
+    List<String> unselectedPrimaryKeys = detail.getPrimaryKey().stream()
+      // if there are no columns specified, it means this primary key will be read/selected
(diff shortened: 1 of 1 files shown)


delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -290,14 +290,29 @@ private TableAssessmentResponse assessTable(SourceTable sourceTable, TableRegist
...
     TableDetail detail = tableRegistry.describeTable(db, table);
+    List<Problem> missingFeatures = new ArrayList<>(detail.getFeatures());
+    List<String> unselectedPrimaryKeys = detail.getPrimaryKey().stream()
+      // if there are no columns specified, it means this primary key will be read/selected
+      .filter(columnWhitelist.isEmpty() ? col -> false : col -> !columnWhitelist.contains(col))
+      .collect(Collectors.toList());
+    if (!unselectedPrimaryKeys.isEmpty()) {
+      missingFeatures.add(
+        new Problem("Missing Primary Key",
+                    String.format("Column(s) '%s' is(are) part of the primary key for table '%s' in database '%s', " +
```
albertshau: it's annoying to write the code, but it's a cleaner experience if only 'is' or 'are' are displayed in the error message, not both. Let's check for the size of unselectedPrimaryKeys and modify the message accordingly.

shifuxu0301: Sure, we can definitely do that.

albertshau (approved):

PR #48, opened 2020-03-31
Add another builder in DMLEvent/DDLEvent which takes existing event obj as parameter.

shifuxu0301:


albertshau (commented):
can you also add a similar one to DDLEvent?

shifuxu0301:
> can you also add a similar one to DDLEvent?

Sure, done.

albertshau (approved):

PR #49, opened 2020-04-01
Fixed the issue that delta app will ignore CREATE DATABASE ddl event during snapshotting

shifuxu0301:
1. put DROP_DATABASE ddl operation as blacklisted by default.
2. Do not ignore CREATE_DATABASE event. 

Tested on local, dataset was successfully created in big query target.

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
+++ b/delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
@@ -170,8 +170,6 @@ public class DeltaWorker extends AbstractWorker {
     Set<DDLOperation> ddlBlacklist = new HashSet<>(config.getDdlBlacklist());
     // targets will not behave properly if they don't get create table events
     ddlBlacklist.remove(DDLOperation.CREATE_TABLE);
-    // blacklist drop database event by default in case that user unintentionally drop database from source
-    ddlBlacklist.add(DDLOperation.DROP_DATABASE);
     Set<SourceTable> expandedTables = config.getTables().stream()
       .filter(t -> assignedTables.contains(new TableId(t.getDatabase(), t.getTable(), t.getSchema())))
       .map(t -> {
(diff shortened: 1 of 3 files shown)


delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -170,6 +170,8 @@ public void initialize(WorkerContext context) throws Exception {
     Set<DDLOperation> ddlBlacklist = new HashSet<>(config.getDdlBlacklist());
     // targets will not behave properly if they don't get create table events
     ddlBlacklist.remove(DDLOperation.CREATE_TABLE);
+    // blacklist drop database event by default in case that user unintentionally drop database from source
+    ddlBlacklist.add(DDLOperation.DROP_DATABASE);
```
albertshau: instead of adding it here, where the user can never change the behavior, it's better to change config.getDdlBlacklist() to return a list with DROP_DATABASE if the value is null. That way the user can overwrite it if they wish.

delta-app/src/main/java/io/cdap/delta/app/QueueingEventEmitter.java
```diff
@@ -88,6 +88,11 @@ private boolean shouldIgnore(DDLEvent event) {
     if (sourceTable != null && sourceTable.getDdlBlacklist().contains(event.getOperation())) {
       return true;
     }
+
+    if (sourceTable == null && event.getOperation() == DDLOperation.CREATE_DATABASE) {
```
albertshau: if the operation is create_database or drop_database. We should only rely on the ddl blacklist for filtering out drop_database

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
shifuxu0301: I see. Will do.

delta-app/src/main/java/io/cdap/delta/app/QueueingEventEmitter.java
shifuxu0301: Sure, will add check for create database and drop database if they are not blacklisted.

albertshau (approved):

PR #51, opened 2020-04-08
CDAP-16552 fix worker to stop correctly when source is failing

albertshau:
Fixed a bug that caused the worker to spin forever in a retry
loop when retrying source failures. It was due to a mis-use of
failsafe's abort logic.

Also added additional error handling for when the event reader
or consumer cannot be stopped for some reason.


shifuxu0301 (approved):
LGTM.

PR #53, opened 2020-04-13
CDAP-16554 make table assessor closeable

albertshau:
Made the TableAssessor closeable in case a plugin needs to
release any resources. Also fixed a bug where the TableRegistry
was not getting closed in the draft service.


shifuxu0301 (approved):
LGTM

PR #59, opened 2020-05-04
CDAP-16737 protect against bad event readers

albertshau:
Changed the failure retry logic so that a new event queue is
used instead of re-using the existing queue. This is to protect
against bad event readers that are not really stopped after they
return from the stop() call. This prevents those types of
readers from adding events to the queue that shouldn't be there,
breaking the event ordering guarantee.

Also added some logging when readers and consumers are starting
and stopping so that the lifecycle of everything is more clear.

Also made the EventEmitter throw InterruptedException, since the
caller really does need to know when it is interrupted in
order to handle it properly and ensure everything shuts down.

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
+++ b/delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
@@ -19,6 +19,7 @@ package io.cdap.delta.app;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
+import io.cdap.cdap.api.Resources;
 import io.cdap.cdap.api.app.ApplicationSpecification;
 import io.cdap.cdap.api.macro.MacroEvaluator;
 import io.cdap.cdap.api.metrics.Metrics;
@@ -74,6 +75,7 @@ public class DeltaWorker extends AbstractWorker {
   private static final Gson GSON = new Gson();
   private static final String GENERATION = "generation";
   private static final String TABLE_ASSIGNMENTS = "table.assignments";
+  private static final String EVENT_QUEUE_SIZE = "event.queue.size";
   private static final Type TABLE_ASSIGNMENTS_TYPE = new TypeToken<Map<Integer, Set<TableId>>>() { }.getType();
 
   private final AtomicBoolean shouldStop;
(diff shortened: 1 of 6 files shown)


delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -186,8 +188,8 @@ public void initialize(WorkerContext context) throws Exception {
     readerDefinition = new EventReaderDefinition(expandedTables,
                                                  config.getDmlBlacklist(),
                                                  config.getDdlBlacklist());
-
-    eventQueue = new ArrayBlockingQueue<>(100);
+    eventQueueSize = Integer.parseInt(context.getRuntimeArguments().getOrDefault(EVENT_QUEUE_SIZE, "10"));
```
shifuxu0301, line 195: should we set the default queue size to 100?

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -394,19 +400,27 @@ private void applyEvent(Sequenced<? extends ChangeEvent> event) throws Exception
   }
 
   private void startFromLastCommit() throws Exception {
-    eventQueue.clear();
```
shifuxu0301, line 397: It seems like the clear() method will atomically removes all of the elements in the queue. Can we have a more detailed example here? I am not too sure what's the difference here?

shifuxu0301: Got it. Thanks for the comments.

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
shifuxu0301: Before merge, can we change the default value to 100, feel like 10 is too small. 
Also, is there any doc for delta app. Maybe it is good to show how to use the runtime argument to set the customized blocking queue size.

shifuxu0301 (approved):
Overall lgtm. have one small comment in for the blocking queue size.

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
albertshau: I'll add more comments, but if we get 100 update events, that means we have roughly 2gb / 100 = 20mb per event, which means roughly 10mb per row since updates contain the data from both before and after. We've gotten questions about being able to handle events of this size, so I think it's better to use a smaller size by default.

In general, I think it's better to set defaults that skew towards reliability, and let the more advanced users adjust for performance. I've also opened https://issues.cask.co/browse/CDAP-16755 to do this based on event size rather than number of events.

albertshau:
@shifuxu0301 I've merged with 10 for now. I don't think the queue size actually matters in most situations (I didn't notice anything in my testing). If you notice a performance hit please let me know.

PR #74, opened 2020-10-26
[CDAP-17361] Fix assessTable to honor columns selection

elfenheart:
JIRA: https://issues.cask.co/browse/CDAP-17361

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/store/DraftService.java
+++ b/delta-app/src/main/java/io/cdap/delta/store/DraftService.java
@@ -196,6 +196,7 @@ public class DraftService {
    * @throws TableNotFoundException if the table does not exist
    * @throws IOException if the table detail could not be read
    * @throws Exception if there was an error creating the table registry
+   * @throws IllegalArgumentException if the table is not selected in the draft
    */
   public TableAssessmentResponse assessTable(DraftId draftId, Configurer configurer, String db, String table)
     throws Exception {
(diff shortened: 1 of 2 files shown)


delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -206,10 +206,15 @@ public TableAssessmentResponse assessTable(DraftId draftId, Configurer configure
     if (target == null) {
       throw new InvalidDraftException("Cannot assess a table without a configured target.");
     }
+    SourceTable selectedTable = deltaConfig.getTables()
+            .stream()
+            .filter(streamTable -> db.equals(streamTable.getDatabase()) && table.equals(streamTable.getTable()))
+            .findAny()
+            .orElse(null);
```
seanzhougoogle: I think we need to handle the case where the selectedTable is null, which means the table in the request cannot be found in the draft.

seanzhougoogle: Also it would be better add the corresponding test case in the same request to verify your changes :)

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -206,10 +206,18 @@ public TableAssessmentResponse assessTable(DraftId draftId, Configurer configure
     if (target == null) {
       throw new InvalidDraftException("Cannot assess a table without a configured target.");
     }
+    SourceTable selectedTable = deltaConfig.getTables()
+            .stream()
```
albertshau: nit: indentation should be 2 spaces:

```
SourceTable selectedTable = deltaConfig.getTables()
  .stream()
  ...
```

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -206,10 +206,18 @@ public TableAssessmentResponse assessTable(DraftId draftId, Configurer configure
...
+            .filter(streamTable -> db.equals(streamTable.getDatabase()) && table.equals(streamTable.getTable()))
+            .findAny()
+            .orElse(null);
+    if (selectedTable == null) {
+      throw new TableNotFoundException(db, table, "Cannot assess table.");
```
albertshau: Doesn't seem like we should hit this case, but should use a better error message anyway. Something like:

table 'xyz' in db 'abc' is not a selected table in the draft.

This also shouldn't be a TableNotFoundException, as that exception is used to denote that the table doesn't exist in the database (some sort of user error). This is more like an IllegalArgumentException, as it indicates we have a bug somewhere in our logic.

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
albertshau: nit: this and the subsequent null check can be replaced with `.orElseThrow(() -> new IllegalArgumentException(...)))`

albertshau (approved):

PR #75, opened 2020-10-30
CDAP-17371 Update replicator app to emit table level metrics.

sagarkapare:


delta-api/src/main/java/io/cdap/delta/api/DeltaTargetContext.java
```diff
@@ -27,17 +27,19 @@
    * Record that a DML operation was processed. Metrics on the number of operations processed are stored in memory
    * and actually written out when {@link #commitOffset(Offset, long)} is called.
    *
+   * @param tableName name of the source table for which count to be incremented
    * @param op type of DML operation
    */
-  void incrementCount(DMLOperation op);
+  void incrementCount(String tableName, DMLOperation op);
```
albertshau: instead of adding an argument, I think it would be better to create a new object that has table name and op as fields. This would make the API more extensible, in case we need add anything in the future.

delta-app/src/main/java/io/cdap/delta/app/DeltaContext.java
```diff
@@ -195,7 +204,7 @@ public void notifyFailed(Throwable cause) {
   }
 
   void clearMetrics() {
-    eventMetrics.clear();
+    tableEventMetrics.clear();
```
albertshau, line 213: this should loop through the EventMetrics and clear each individual one.

albertshau (commented):
any changes needed for unit tests?

delta-api/src/main/java/io/cdap/delta/api/DeltaTargetContext.java
sagarkapare: Refactored code to add DMLOperation and DDLOperation as classes with table name in them. Moved existing operation type enums into member variables of DMLOperation and DDLOperation. 

This method now takes DMLOperation.

delta-app/src/main/java/io/cdap/delta/app/DeltaContext.java
sagarkapare: Any reason for looping through it?

sagarkapare:
Manually tested the functionality.
1. Sample curl request for aggregate:
```
curl -v -X POST "localhost:11015/v3/metrics/query?tag=namespace:default&tag=app:Replicator&tag=worker:DeltaWorker&tag=run:*&tag=instance:*&tag=ent:pet&metric=user.dml.insert&start=now-1h&end=now&aggregate=true" | python -m json.tool
```

Corresponding response:

```
{
    "endTime": 1604472716,
    "resolution": "2147483647s",
    "series": [
        {
            "data": [
                {
                    "time": 0,
                    "value": 7
                }
            ],
            "grouping": {},
            "metricName": "user.dml.insert"
        }
    ],
    "startTime": 0
}
```

2. Sample curl request with groupBy:

```
curl -v -X POST "localhost:11015/v3/metrics/query?tag=namespace:default&tag=app:Replicator&tag=worker:DeltaWorker&tag=run:*&tag=instance:*&groupBy=ent&metric=user.dml.insert&start=now-1h&end=now&aggregate=true" | python -m json.tool
```

Corresponding response:

```
{
    "endTime": 1604472849,
    "resolution": "2147483647s",
    "series": [
        {
            "data": [
                {
                    "time": 0,
                    "value": 7
                }
            ],
            "grouping": {
                "null": "pet"
            },
            "metricName": "user.dml.insert"
        },
        {
            "data": [
                {
                    "time": 0,
                    "value": 1
                }
            ],
            "grouping": {
                "null": "MyGuests"
            },
            "metricName": "user.dml.insert"
        }
    ],
    "startTime": 0
}
```
In the response above grouping key is "null" rather than "ent", something probably need to fix in cdap platform (looks like it needs to be added to this map - https://github.com/cdapio/cdap/blob/develop/cdap-watchdog/src/main/java/io/cdap/cdap/metrics/query/MetricsQueryHelper.java#L84).

delta-api/src/main/java/io/cdap/delta/api/DeltaTargetContext.java
```diff
@@ -27,15 +27,15 @@
    * Record that a DML operation was processed. Metrics on the number of operations processed are stored in memory
    * and actually written out when {@link #commitOffset(Offset, long)} is called.
    *
-   * @param op type of DML operation
+   * @param op represents DML operation
```
seanzhougoogle: nit: no need to add "represents" , implicitly the comments here is to explain what does the param represent. Also you commented in a different way for DDL Operation param below (which I prefer that way), let's keep them consistent :)

delta-api/src/main/java/io/cdap/delta/api/DeltaTargetContext.java
```diff
@@ -47,7 +47,8 @@
    * always have the same corresponding sequence number.
    *
    * Also writes out metrics on the number of DML and DDL operations that were applied since the last successful
-   * commit, as recorded by calls to {@link #incrementCount(DDLOperation)} and {@link #incrementCount(DMLOperation)}.
+   * commit, as recorded by calls to {@link #incrementCount(String, DDLOperation.Type)} and
```
seanzhougoogle: There is no "incrementCount(String, DDLOperation.Type)" , should it be incrementCount(DDLOperation.Type) ? a typo?

delta-api/src/main/java/io/cdap/delta/api/DDLEvent.java
```diff
@@ -28,14 +28,14 @@
  * A DDL change event.
  */
 public class DDLEvent extends ChangeEvent {
-  private final DDLOperation operation;
+  private final DDLOperation.Type operation;
   private final Schema schema;
   private final String database;
   private final String prevTable;
   private final String table;
```
albertshau, line 35: if the table name is in the Operation, it shouldn't also be in the event.

delta-app/src/main/java/io/cdap/delta/app/DeltaContext.java
```diff
@@ -73,18 +75,31 @@
...
   }
 
   @Override
   public void commitOffset(Offset offset, long sequenceNumber) throws IOException {
     stateStore.writeOffset(id, new OffsetAndSequence(offset, sequenceNumber));
-    eventMetrics.emitMetrics();
+    for (EventMetrics eventMetrics : tableEventMetrics.values()) {
+      eventMetrics.emitMetrics();
+    }
+    tableEventMetrics.clear();
```
seanzhougoogle: nit: shall we call clearMetrics() here ? so that if you have implementation changes in clearMetrics, it will be auto applied here.

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -159,10 +159,9 @@ public void initialize(WorkerContext context) throws Exception {
 
     Path path = new Path(offsetBasePath);
     StateStore stateStore = StateStore.from(path);
-    EventMetrics eventMetrics = new EventMetrics(metrics, targetName);
     PipelineStateService stateService = new PipelineStateService(id, stateStore);
     stateService.load();
-    deltaContext = new DeltaContext(id, context.getRunId().getId(), metrics, stateStore, context, eventMetrics,
+    deltaContext = new DeltaContext(id, context.getRunId().getId(), metrics, stateStore, context,
```
seanzhougoogle, line 164: so we drop the information of targetName ? (the prefix)

seanzhougoogle (approved):
Nice job , Sagar :)

delta-api/src/main/java/io/cdap/delta/api/DDLOperation.java
```diff
@@ -16,15 +16,59 @@
...
+    CREATE_DATABASE,
+    DROP_DATABASE,
+    CREATE_TABLE,
+    DROP_TABLE,
+    TRUNCATE_TABLE,
+    ALTER_TABLE,
+    RENAME_TABLE
+  }
+
+  private final String tableName;
```
albertshau, line 39: I think it makes sense for this to also have the prevTableName field, as a RENAME_TABLE operation will need it. I would assume we need to update the ddl count for both the previous table name and the new one?

delta-app/src/main/java/io/cdap/delta/app/DeltaContext.java
albertshau: actually, not needed if EventMetrics.clear() doesn't do anything except clear in-memory things. Can ignore my previous comment

delta-api/src/main/java/io/cdap/delta/api/DDLEvent.java
sagarkapare: yeah noticed it earlier.. will remove it from `DDLEvent` and `DMLEvent` classes as a followup PR.

delta-api/src/main/java/io/cdap/delta/api/DDLOperation.java
sagarkapare: added prevTableName for rename operation.

delta-app/src/main/java/io/cdap/delta/app/DeltaContext.java
sagarkapare: makes sense 👍 fixed.

delta-api/src/main/java/io/cdap/delta/api/DeltaTargetContext.java
sagarkapare: fixed.

delta-api/src/main/java/io/cdap/delta/api/DeltaTargetContext.java
sagarkapare: ah good catch! forgot to modify this comment after i modified the method signature. thanks!

albertshau (approved):

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
sagarkapare: `targetName` in  metrics for dml and ddl events seems redundant. If we want we can emit separate metric for counting number of records inserted in the target, however there is no use case for that currently.

sagarkapare:
Thank you @albertshau and @seanzhougoogle for the comments. Squashing the commits and merging.

PR #76, opened 2020-11-05
CDAP-17371 Removed table name from DDLEvent and DMLEvent classes since they are now part of DMLOperation and DDLOperation.

sagarkapare:


delta-api/src/main/java/io/cdap/delta/api/DDLEvent.java
```diff
@@ -160,7 +145,13 @@ public Builder setPrimaryKey(List<String> primaryKey) {
     }
 
     public DDLEvent build() {
-      return new DDLEvent(offset, operation, schema, database, prevTable, table, primaryKey, isSnapshot);
+      DDLOperation ddlOperation;
+      if (operation == DDLOperation.Type.RENAME_TABLE) {
+        ddlOperation = DDLOperation.createRenameTableOperation(prevTable, table);
```
seanzhougoogle, line 150: why the DDLEvent build need to take responsibility for build DDL operation instance ? Should it be the responsibility of a DDL operation builder ?

sagarkapare: `DDLEvent.build()` method is creating instance of `DDLEvent`. `DDLOperation` is member of that instance. So it needs to create DDLOperation instance here. 

I think what you are suggesting here is why don't we have builder similar to `DDLEvent` for `DDLOperation` class. If we had that builder, code here would still look like following
```
ddlOperation = DDLOperation.builder().setPreviousTable(prevTable).setTableName(tableName).build();
```
We don't need builder for `DDLOperation` since it does not fit well yet as number of members of that class are limited.

Let me know if you had something else in mind.

seanzhougoogle: Thank you for the detailed explanation :) 
actually I was thinking , we should let client of DDLEvent to pass in a DDLOperation instance or DDLOperation builder instead of all the  members of the DDLOperation. Just like what proto builder did. If a proto contains a sub proto, its builder doesn't accept the member of a sub proto , instead it accepts an instance of the sub proto or sub proto's builder. The advantage of this way is that it's scalable and make the responsibility clear. 

However I think it's fine here , since the use case here is simple enough :)

seanzhougoogle (approved):

PR #77, opened 2020-11-11
Fixed test cases by removing the target name from metric.

sagarkapare:
As a part of fix for bug https://issues.cask.co/browse/CDAP-17371, we removed the target names from the metric names being emitted. Similar changes are required in the test case too.


sagarkapare:
Build is green here - https://builds.cask.co/browse/DEL-BT75-1

MEseifan (approved):

PR #79, opened 2020-11-17
CDAP-17407 Add API for the source to be able to provide additional properties.

sagarkapare:
JIRA - https://issues.cask.co/browse/CDAP-17407


delta-api/src/main/java/io/cdap/delta/api/SourceProperties.java
```diff
@@ -0,0 +1,120 @@
...
+  public static class Builder {
+    private Ordering ordering;
+    private RowIdKey rowIdKey;
+
+    public Builder() {
+      ordering = Ordering.ORDERED;
+    }
+
+    /**
+     * Specify the ordering of the events generated by source.
```
seanzhougoogle: nit: Sets (to keep it consistent with comments of other set method)

delta-api/src/main/java/io/cdap/delta/api/SourceProperties.java
```diff
@@ -0,0 +1,120 @@
...
+     * @return this builder
+     */
+    public SourceProperties.Builder setOrdering(Ordering ordering) {
+      this.ordering = ordering;
+      return this;
+    }
+
+    /**
+     * Sets the row id key which will uniquely identify the row in the table.
+     * Set the key which is used to uniquely identify a particular row in the database.
```
seanzhougoogle: this line is kind of dup with above line. Should we just emphasize that it should not be changeable for any DML operation unless deleted ?

delta-api/src/main/java/io/cdap/delta/api/RowIdKey.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Represent row id keys which is uniquely determines the row in a database.
+ */
+public class RowIdKey {
+  private final List<String> rowIdKeys;
```
seanzhougoogle, line 28: Does the order of the keys matter ?

delta-app/src/test/java/io/cdap/delta/api/SourcePropertiesTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.google.gson.Gson;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Arrays;
+
+public class SourcePropertiesTest {
+  @Test
+  public void testSerDe() throws Exception {
```
seanzhougoogle, line 27: do you want to add some test case to test your validations ? (e.g. ordered with row id key  and un ordered without row id key )

seanzhougoogle (approved):
Thank you Sagar :)

delta-api/src/main/java/io/cdap/delta/api/SourceProperties.java
sagarkapare: fixed.

delta-api/src/main/java/io/cdap/delta/api/SourceProperties.java
sagarkapare: updated the description. ptal again!

delta-api/src/main/java/io/cdap/delta/api/RowIdKey.java
sagarkapare: Ordering of the keys matter from database perspective. Specifically for primary keys - https://stackoverflow.com/questions/16713233/why-does-primary-key-order-matter. Also looked through debezium code base and they are using List rather than Set for representing different keys.

seanzhougoogle (approved):

sagarkapare:
Build is green - https://builds.cask.co/browse/DEL-BT77-5. Squashing and merging

PR #81, opened 2020-12-01
CDAP-17407 Update DMLEvent API so that source can set the source_timestamp.

sagarkapare:
Also removed the RowIdKey from the SourceProperties and setting it through DMLEvent API itself. Reason for this removal is row id in general in the context of CDC is a metadata of a row rather than the actual data itself. For this metadata, we can simply assume the column name as `_row_id` in the app and let user provide its value. Also made it as single String rather than List, as based on the Oracle use case row_id is represented by single field only.


delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -31,17 +31,21 @@
   private final StructuredRecord previousRow;
   private final String transactionId;
   private final long ingestTimestampMillis;
+  private final Long sourceTimestampMillis;
+  private final String rowId;
 
   private DMLEvent(Offset offset, String database, DMLOperation operation, StructuredRecord row,
                    @Nullable StructuredRecord previousRow, @Nullable String transactionId, long ingestTimestampMillis,
-                   boolean isSnapshot) {
+                   @Nullable Long sourceTimestampMillis, boolean isSnapshot, String rowId) {
```
seanzhougoogle: `@Nullable` String rowId ?

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -31,17 +31,21 @@
   private final StructuredRecord previousRow;
   private final String transactionId;
   private final long ingestTimestampMillis;
+  private final Long sourceTimestampMillis;
```
seanzhougoogle: I guess the reason you are not using a primitive long here is because you want to make it nullable , right ?
so how does a default value 0 hurt here for a primitive ?

seanzhougoogle (approved):
Thank you Sagar :)

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -166,9 +185,19 @@ public Builder setIngestTimestamp(long ingestTimestampMillis) {
       return this;
     }
 
+    public Builder setSourceTimestamp(long sourceTimestampMillis) {
```
seanzhougoogle: shall we mention in comments that for unordered events , source timestamp must be set ?

Or even put an ordering attribute in the event so that we can validate that this property being set for unordered events in the build() method ?

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
sagarkapare: fixed.

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
sagarkapare: wanted to avoid additional storage cost with default value if any.. besides now I am using null check to assign the sequence number. please take a look again thanks!

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
sagarkapare: Added comment. We don't need to put ordering attribute on each event as its property of the source. We can possibly look into the validation as a part of separate bug.

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
seanzhougoogle: LGTM, thank you , Sagar :)

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
seanzhougoogle: sg

sagarkapare:
Build is green here - https://builds.cask.co/browse/DEL-BT80-5 squashing and merging.

PR #82, opened 2020-12-03
CDAP-17411 return delta pipeline id in delta context

seanzhougoogle:


seanzhougoogle:
The reason for this change is because some plugins may need this id to determine whether some resources have already been created. Those resources may be created once when first time pipeline is run. Because generation is generated based on the timestamp when pipeline is deployed.  This id is kind of global unique,unless there are two pipeline with same namespace and name deployed at the same time in two CDAP instances.

delta-app/src/main/java/io/cdap/delta/app/DeltaPipelineId.java
```diff
@@ -42,4 +42,9 @@ public String getApp() {
   public long getGeneration() {
     return generation;
   }
+
+  @Override
+  public String toString() {
+    return namespace + "-" + app + "-" + generation;
```
albertshau: Think it would be better to move this whole class to the delta-api module and have getPipelineId() return this class instead of a stringified version.

seanzhougoogle: thanks for the suggestion:) Done :)

albertshau (approved):

PR #83, opened 2020-12-03
CDAP-17437 expose namesapce, name and generation in Configurer

seanzhougoogle:
TableAssessor may create some temporary resources. With draft id as prefix of the resource ID, it would be easy to detect/cleanup those dangling resources

### DIFF
--- a/delta-api/src/main/java/io/cdap/delta/api/Configurer.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/Configurer.java
@@ -18,9 +18,19 @@ package io.cdap.delta.api;
 
 import io.cdap.cdap.api.plugin.PluginConfigurer;
 
+import javax.annotation.Nullable;
+
 /**
  * Configures a data transfer.
  */
 public interface Configurer extends PluginConfigurer {
 
+  @Nullable
+  String getNamespace();
+
+  @Nullable
+  String getName();
+
+  @Nullable
+  Long getGeneration();
 }
--- a/delta-api/src/main/java/io/cdap/delta/api/Configurer.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/Configurer.java
@@ -30,7 +30,4 @@ public interface Configurer extends PluginConfigurer {
 
   @Nullable
   String getName();
-
-  @Nullable
-  Long getGeneration();
 }
--- a/delta-api/src/main/java/io/cdap/delta/api/assessment/TableAssessorSupplier.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/assessment/TableAssessorSupplier.java
@@ -32,20 +32,5 @@ public interface TableAssessorSupplier<T> {
    * @return a table assessor
    * @throws Exception if the table assessor could not be created
    */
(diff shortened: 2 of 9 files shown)


delta-api/src/main/java/io/cdap/delta/api/assessment/TableAssessorSupplier.java
```diff
@@ -33,4 +33,17 @@
...
+  /**
+   * Create a table assessor that will check if there will be potential problems replicating a table.
+   *
+   * @param configurer configurer used to instantiate plugins
+   * @param tempResourceIdPrefix the prefix of the id for some temporary resources created by Assessor. With the
+   *                             prefix, it will be easy to detect/cleanup dangling resources
+   * @return a table assessor
+   * @throws Exception if the table assessor could not be created
+   */
+  default TableAssessor<T> createTableAssessor(Configurer configurer, String tempResourceIdPrefix) throws Exception {
```
albertshau: instead of adding a new method, it would be better to expose methods like getNamespace(), getName() in the Configurer interface

seanzhougoogle: thanks for the suggestion :) done :)

delta-api/src/main/java/io/cdap/delta/api/Configurer.java
```diff
@@ -18,9 +18,19 @@
...
 import io.cdap.cdap.api.plugin.PluginConfigurer;
 
+import javax.annotation.Nullable;
+
 /**
  * Configures a data transfer.
  */
 public interface Configurer extends PluginConfigurer {
 
+  @Nullable
```
albertshau, line 28: Are these ever null? It seems like they shouldn't be nullable

delta-api/src/main/java/io/cdap/delta/api/Configurer.java
```diff
@@ -18,9 +18,19 @@
...
+  String getNamespace();
+
+  @Nullable
+  String getName();
+
+  @Nullable
+  Long getGeneration();
```
albertshau: Do we need the namespace generation? This might be confusing to have since it's different than the pipeline id generation.

If we need it, I think it would be better to move Namespace to delta-api and make getNamespace() return that class, that way it's more clear that it's the namespace generation.

delta-api/src/main/java/io/cdap/delta/api/Configurer.java
seanzhougoogle: Mainly because I don't see a way to get the namespace and name here:
https://github.com/data-integrations/delta/blob/develop/delta-app/src/main/java/io/cdap/delta/app/DeltaApp.java#L50

delta-api/src/main/java/io/cdap/delta/api/Configurer.java
seanzhougoogle: I was thinking generation was one of the identifier of namespace. If that's not the case, I don't think we need it.

seanzhougoogle:
discussed with Albert offline. We can just use "_cdap" or "_temp" as the prefix , so this PR is no longer needed.

PR #88, opened 2020-12-08
[CDAP-17371] Emit metric for bytes processed by the replicator pipeline.

sagarkapare:
1. Add method to compute the size of StructuredRecord based on the field types. We can make certain assumptions in the replicator app that the schema is not nested.
2. Emit metric `dml.data.processed.bytes` at pipeline level.

### DIFF
--- a/delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
@@ -16,8 +16,15 @@
 
 package io.cdap.delta.api;
 
+import io.cdap.cdap.api.common.Bytes;
 import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.api.data.schema.Schema;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
 import java.util.Objects;
 import javax.annotation.Nullable;
 
@@ -139,6 +146,7 @@ public class DMLEvent extends ChangeEvent {
       this.transactionId = event.getTransactionId();
       this.ingestTimestampMillis = event.getIngestTimestampMillis();
       this.isSnapshot = event.isSnapshot();
+      this.sourceTimestampMillis = event.getSourceTimestampMillis();
     }
 
     public Builder setOperationType(DMLOperation.Type operationType) {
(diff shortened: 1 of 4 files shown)


delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -182,8 +186,53 @@ public Builder setRowId(String rowId) {
     }
 
     public DMLEvent build() {
-      return new DMLEvent(offset, database, new DMLOperation(table, operationType, ingestTimestampMillis), row,
-                          previousRow, transactionId, ingestTimestampMillis, sourceTimestampMillis, isSnapshot, rowId);
+      int sizeInBytes = (operationType == DMLOperation.Type.INSERT ||
```
seanzhougoogle, line 193: should we calculate the size of Delete from previousRow ?
it seems not reasonable for size 0 for delete operation

seanzhougoogle: for example, if there are all delete operations, our metrics will look like we didn't process any data

sagarkapare: It's the amount of bytes we write to the BigQuery. Since for DELETE we don't write anything @bdmogal suggested to count it as 0.

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -182,8 +186,53 @@ public Builder setRowId(String rowId) {
...
+            break;
+          case STRING:
+            String strValue = row.get(field.getName());
+            sizeInBytes += strValue.getBytes(StandardCharsets.UTF_8).length;
+            break;
+          default:
+            break;
+        }
+      }
+      return sizeInBytes + numBooleans / 8;
```
seanzhougoogle: although booleans can be represented by bit but multiple booleans won't be merged into one byte. one boolean still occupies one byte.

albertshau: +1, should be able to just do sizeInBytes++ for the boolean case

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -182,8 +186,53 @@ public Builder setRowId(String rowId) {
...
+          case FLOAT:
+          case ENUM:
+            sizeInBytes += 4;
+            break;
+          case LONG:
+          case DOUBLE:
+            sizeInBytes += 8;
+            break;
+          case BYTES:
+            byte[] bytes = row.get(field.getName());
```
albertshau: we allow ByteBuffer or byte[] for byte fields in StructuredRecord

sagarkapare: fixed.

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
sagarkapare: done.

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -182,8 +190,56 @@ public Builder setRowId(String rowId) {
...
+          case FLOAT:
+          case ENUM:
+            sizeInBytes += 4;
+            break;
+          case LONG:
+          case DOUBLE:
+            sizeInBytes += 8;
+            break;
+          case BYTES:
+            Object value = row.get(field.getName());
```
albertshau: nit: can have this variable before the null check (line 210) to do just a single get instead of 2.

albertshau (approved):

sagarkapare:
Build is green here squashing and merging - https://builds.cask.co/browse/DEL-BT88

PR #89, opened 2020-12-09
CDAP-17418 add schema field to DML/DDL event

seanzhougoogle:


delta-api/src/main/java/io/cdap/delta/api/DDLOperation.java
```diff
@@ -36,12 +36,17 @@
     RENAME_TABLE
   }
 
+  private final String schema;
```
sagarkapare: can we move database name to this class too? from DMLEvent and DDLEvent? So that table identifier - databasename, schema and table name at the same place?

delta-api/src/main/java/io/cdap/delta/api/DDLEvent.java
```diff
@@ -30,15 +30,13 @@
 public class DDLEvent extends ChangeEvent {
   private final DDLOperation operation;
   private final Schema schema;
-  private final String database;
   private final List<String> primaryKey;
 
-  private DDLEvent(Offset offset, String database, DDLOperation operation, @Nullable Schema schema,
+  private DDLEvent(Offset offset, DDLOperation operation, @Nullable Schema schema,
```
sagarkapare, line 35: schema is not required in this class since it will come from DDLOperation.

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -139,8 +138,9 @@ private Builder() { }
     private Builder(DMLEvent event) {
       this.offset = event.getOffset();
       this.operationType = event.getOperation().getType();
-      this.database = event.getDatabase();
-      this.table = event.getOperation().getTableName();
+      this.databaseName = event.getDatabase();
```
sagarkapare: event.getOperation().getDatabaseName()?

delta-api/src/main/java/io/cdap/delta/api/DDLOperation.java
seanzhougoogle: done:) PTAL :)

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -193,9 +198,9 @@ public DMLEvent build() {
       int sizeInBytes = (operationType == DMLOperation.Type.INSERT ||
         operationType == DMLOperation.Type.UPDATE) ? computeSizeInBytes(row) : 0;
 
-      return new DMLEvent(offset, database, new DMLOperation(table, operationType, ingestTimestampMillis, sizeInBytes),
-                          row, previousRow, transactionId, ingestTimestampMillis, sourceTimestampMillis, isSnapshot,
-                          rowId);
+      return new DMLEvent(offset, databaseName,
```
sagarkapare: We can remove databaseName from DMLEvent since its already part of DMLOperation.

delta-api/src/main/java/io/cdap/delta/api/DMLOperation.java
```diff
@@ -17,33 +17,49 @@
...
 
 import java.util.Objects;
+import javax.annotation.Nullable;
 
 /**
  * Represents a DML Operation.
  */
 public class DMLOperation {
 
+
```
sagarkapare: nit: remove empty lines.

sagarkapare (commented):
Few minor comments. Also please pull latest from develop to check if new api change is required at other places.

delta-api/src/main/java/io/cdap/delta/api/DDLEvent.java
seanzhougoogle: This is table schema, not schema name. Do we want to move the table schema to DDLOpration too ?

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
seanzhougoogle: done :) removed database from DMLEvent too:)

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
seanzhougoogle: done :)

delta-api/src/main/java/io/cdap/delta/api/DMLOperation.java
seanzhougoogle: done:)

delta-api/src/main/java/io/cdap/delta/api/DDLEvent.java
sagarkapare: ah its ok then. thanks for clarification.

sagarkapare (approved):

PR #95, opened 2020-12-11
CDAP-17462 Add supportRowId to source properties and chagne eventsOrdered to boolean

seanzhougoogle:
CDAP-17462 Add supportRowId to source properties and chagne eventsOrdered to boolean


delta-api/src/main/java/io/cdap/delta/api/SourceProperties.java
```diff
@@ -26,19 +26,15 @@
   /**
    * Enum describing ordering of events.
```
sagarkapare: Remove this comment too?

sagarkapare (approved):

delta-api/src/main/java/io/cdap/delta/api/SourceProperties.java
```diff
@@ -26,19 +26,15 @@
...
    */
-  public enum Ordering {
-    ORDERED,
-    UN_ORDERED
-  }
 
-  private final Ordering ordering;
+  private final boolean supportRowId;
```
sagarkapare: nit: should we rename this to `rowIdSupported` and corresponding getter can be `isRowIdSupported` and setter `setRowIdSupported`?

delta-api/src/main/java/io/cdap/delta/api/SourceProperties.java
seanzhougoogle: done:)

delta-api/src/main/java/io/cdap/delta/api/SourceProperties.java
seanzhougoogle: done :)

PR #100, opened 2020-12-16
[CDAP-17478] assessPipeline count to ignore partially supported colum…

elfenheart:
…ns that are suppressed

JIRA: https://cdap.atlassian.net/browse/CDAP-17478

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/store/DraftService.java
+++ b/delta-app/src/main/java/io/cdap/delta/store/DraftService.java
@@ -280,7 +280,7 @@ public class DraftService {
           TableAssessmentResponse assessment = assessTable(sourceTable, tableRegistry, sourceTableAssessor,
                                                            targetTableAssessor);
           missingFeatures.addAll(assessment.getFeatures());
-          tableAssessments.add(summarize(db, sourceTable.getSchema(), table, assessment));
+          tableAssessments.add(summarize(db, sourceTable.getSchema(), sourceTable, assessment));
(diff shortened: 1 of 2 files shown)


delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -473,11 +473,19 @@ private TableSummaryAssessment summarize(String db, @Nullable String schema, Str
...
         numUnsupported++;
       } else if (columnAssessment.getSupport() == ColumnSupport.PARTIAL) {
-        numPartial++;
+        Set<SourceColumn> columns = table.getColumns();
+        SourceColumn existingColumn = columns.stream()
+          .filter(column -> columnAssessment.getSourceName().equals(column.getName()))
+          .findAny()
+          .orElse(null);
+
+        if (existingColumn == null || !existingColumn.isSuppressWarning()) {
```
seanzhougoogle, line 482: in which case it will be null and the partial num should increase ?

elfenheart: column selection is not required, so the array of column selection can be empty, in which case the column is supposed to be replicated

seanzhougoogle: got it. LGTM

seanzhougoogle (approved):

elfenheart:
Green build: https://builds.cask.co/browse/DEL-BT100-2

PR #106, opened 2020-12-23
add schema to TableNotFoundException

seanzhougoogle:


delta-api/src/main/java/io/cdap/delta/api/assessment/TableNotFoundException.java
```diff
@@ -25,7 +25,19 @@ public TableNotFoundException(String database, String table, String errorMessage
     super(String.format("Table '%s' in database '%s' was not found. %s", table, database, errorMessage));
```
sagarkapare: This can just call `this(database, null, table, errorMessage)` to avoid duplicate message creation.

delta-api/src/main/java/io/cdap/delta/api/assessment/TableNotFoundException.java
```diff
@@ -25,7 +25,19 @@ public TableNotFoundException(String database, String table, String errorMessage
...
   }
 
+  public TableNotFoundException(String database, String schema, String table, String errorMessage) {
+    super(String.format("Table '%s' in database '%s' and schema '%s' was not found. %s", table, database, schema,
```
sagarkapare: We can form a message in a variable here depending on the schema is null or not and then use that variable to call super.

delta-api/src/main/java/io/cdap/delta/api/assessment/TableNotFoundException.java
```diff
@@ -25,7 +25,19 @@ public TableNotFoundException(String database, String table, String errorMessage
...
+      errorMessage));
+  }
+
   public TableNotFoundException(String database, String table, String errorMessage, Throwable cause) {
     super(String.format("Table '%s' in database '%s' was not found. %s", table, database, errorMessage), cause);
```
sagarkapare: Also better call the variant `this(database, schema, table, errorMessage, cause)` from here.

sagarkapare (commented):
Minor nits.

delta-api/src/main/java/io/cdap/delta/api/assessment/TableNotFoundException.java
seanzhougoogle: Actually I did that on purpose , because the schema appears in the middle of the error message
to construct the error message based on whether the schema is null will make the code clumsy and not that readable

seanzhougoogle: for example, if we do below
```
if (schema != null) 
  errorMessage = String.format("Table '%s' in database '%s' and schema '%s' was not found. %s", table, database, schema, errorMessage)
else 
 errorMessage = String.format("Table '%s' in database '%s' was not found. %s", table, database, errorMessage)
```
Then it has no difference from my current implementation

seanzhougoogle: If we want to share the common part , then assembling messages will make code clumsy and less readable :
```
message = String.format("Table '%s' in database '%s' ", table, database);
if (schema != null) {
   message = String.format ("%s and schema '%s'" , message, schema);
}
message = String.format("%s  was not found. %s", message, errorMessage);
```
What do you think ?

delta-api/src/main/java/io/cdap/delta/api/assessment/TableNotFoundException.java
seanzhougoogle: My major concern is that assembling the message based on the schema will make the code clumsy and less readable. because the schema information appears in the middle of the message.

seanzhougoogle: also call to super must be the first statement in the constructor.

delta-api/src/main/java/io/cdap/delta/api/assessment/TableNotFoundException.java
sagarkapare: It is not a big deal here but in general since `TableNotFoundException` is a part of API, it is easy for client of it to call the version (database, schema, table) with schema as `null` by mistake. API does not restrict client from doing so. To make it strict, we can check if schema is  `null` and throw an exception but that would be overkill for this case. Another way to make it stricter is implementation can handle `null` correctly. 

We can do something like following right?
```
String schemaInfo = schema == null ? "" : String.format(" and schema %s ", schema);
String.format("Table '%s' in database '%s'%s was not found.", table, database, schemaInfo)
```

Also this avoids un-necessary code duplication.

seanzhougoogle: sg, done :)

sagarkapare (approved):

PR #107, opened 2020-12-23
CDAP-17371 Emit metric for error as soon as it is received from target.

sagarkapare:
Currently we emit all metrics once offset is committed by target in the replicator app. However it is possible in some cases that pipeline keep generating errors. In those cases error metic will not be published since commit offset will never get called. Ideally error should be emitted as soon as its received from target since its independent on committing offset.

### DIFF
--- a/delta-test/src/main/java/io/cdap/delta/test/mock/MockErrorTarget.java
+++ b/delta-test/src/main/java/io/cdap/delta/test/mock/MockErrorTarget.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


delta-test/src/main/java/io/cdap/delta/test/mock/MockErrorTarget.java
```diff
@@ -0,0 +1,102 @@
...
+import io.cdap.delta.api.assessment.TableAssessment;
+import io.cdap.delta.api.assessment.TableAssessor;
+import io.cdap.delta.proto.Artifact;
+import io.cdap.delta.proto.Plugin;
+
+import java.util.Collections;
+import java.util.HashMap;
+
+/**
+ * Mock target used in unit tests. It only generate error metric.
```
seanzhougoogle: nit: "It only sets errors to context"

Actually Target just set errors to context and it's context decide whether to emit the error metric immediately once error is set.

seanzhougoogle (approved):

delta-test/src/main/java/io/cdap/delta/test/mock/MockErrorTarget.java
sagarkapare: fixed.

PR #110, opened 2020-12-30
Add source property ordering

seanzhougoogle:
Why:

Target will need this property to decide whether to use source time stamp to decide the order of the events. for example if an event happening earlier came later, we should ignore this event. This can be done by filtering the events in staging table by its `event source timestamp >= latest merged source timestamp`

Why can't we use source time stamp as sequence number and use `event sequence number >= latest merged sequence number` to filter the events in staging table:

Because sequence number is used for identify whether the event has already been merged to target table (so that no dup event will be merged to target table). Even through we have below merge condition :

https://github.com/data-integrations/bigquery-delta-plugins/blob/7a44e16cd2880f08ab70f0820f2fd13ef92e7630/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java#L893

but INSERT event won't have `_before_` value, thus duplicate insert events will be merged to target table if we have `=` in the filter. And it's possible that multiple events have same source timestamp  ( for a snapshot all the events have same source timestamp


delta-app/src/main/java/io/cdap/delta/app/QueueingEventEmitter.java
```diff
@@ -64,8 +64,7 @@ public void emit(DDLEvent event) {
     }
 
     try {
-      eventQueue.put(new Sequenced<>(event, event.getSourceTimestampMillis() == null ? ++sequenceNumber
-        : event.getSourceTimestampMillis()));
+      eventQueue.put(new Sequenced<>(event, ++sequenceNumber));
```
sagarkapare, line 67: Same change required for `emit(DMLEvent event)` method too on line 82

seanzhougoogle: good catch :) done :)

sagarkapare (approved):

PR #116, opened 2021-01-08
Add capability annotation for delta app and enable capability for tests

greeshmaswaminathan:
- Annotate delta app
- Enable capability for tests so app deployment works


delta-test/src/main/java/io/cdap/delta/test/DeltaPipelineTestBase.java
```diff
@@ -61,4 +68,13 @@ protected static void setupArtifacts(Class<?> appClass) throws Exception {
...
                       MockSource.class, MockTarget.class, MockErrorTarget.class, FailureTarget.class);
   }
+
+  protected static void enableCapability() throws IOException, InterruptedException {
+    String capabilityFolder = getConfiguration().get("capability.config.dir");
+    String cdcFile = "cdc.json";
+    URL url = DeltaPipelineTestBase.class.getClassLoader().getResource(cdcFile);
+    File outFile = new File(capabilityFolder, cdcFile);
+    ByteStreams.copy(Resources.newInputStreamSupplier(url), Files.newOutputStreamSupplier(outFile));
+    Thread.sleep(TimeUnit.MINUTES.toMillis(1));
```
pandyajay10: why do we need to add a sleep here? If we need to add this, can you add a comment inline for this?

pandyajay10 (approved):
Please also get Albert's approval for this one.

delta-test/src/main/java/io/cdap/delta/test/DeltaPipelineTestBase.java
greeshmaswaminathan: Created a bug for fixing this in a generic way. https://cdap.atlassian.net/browse/CDAP-17562

albertshau (approved):

greeshmaswaminathan:
image

This commit has dependency on the change from this PR in cdap - https://github.com/cdapio/cdap/pull/13056

PR #117, opened 2021-01-11
don't remove '\' when evaluate Macro

seanzhougoogle:
what's wrong :

'\' is removed when evaluate macro in service account key, because we enabled the "escape" and below code will remove the '\' :
https://github.com/cdapio/cdap/blob/7dade6fc7376945673913714dea9bbbe79aff2f6/cdap-app-fabric/src/main/java/io/cdap/cdap/internal/app/runtime/plugin/MacroParser.java#L244

solution:
should disable "escape" for delta , it's not correct for service account keys

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
+++ b/delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
@@ -20,6 +20,8 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonSyntaxException;
 import io.cdap.cdap.api.NamespaceSummary;
+import io.cdap.cdap.api.macro.MacroEvaluator;
+import io.cdap.cdap.api.macro.MacroParserOptions;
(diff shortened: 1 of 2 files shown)


delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -309,7 +313,15 @@ private void respond(String namespaceName, HttpServiceResponder responder, Names
     }
 
     try {
-      endpoint.respond(new DraftService(context, new SystemServicePropertyEvaluator(context)), namespace);
```
sagarkapare: Can we `setEscaping('false')` here itself - https://github.com/data-integrations/delta/blob/develop/delta-app/src/main/java/io/cdap/delta/store/SystemServicePropertyEvaluator.java#L32

seanzhougoogle: good catch :). I was thinking SystemServicePropertyEvaluator was in cdap core.

sagarkapare (approved):

PR #122, opened 2021-01-26
CDAP-16755 throttle event queue by event size

seanzhougoogle:


delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -78,6 +78,11 @@ public String getRowId() {
     return rowId;
   }
 
+  public int getSizeInBytes() {
+    // the size of all other fields are fixed, only return the dynamic size of row and previousRow
+    return operation.getSizeInBytes() + computeSizeInBytes(previousRow);
```
albertshau: would be good to have a private variable that is set lazily. If null, compute this number and set it. Otherwise, return the value. That way we prevent recomputing on each call

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
```diff
@@ -0,0 +1,179 @@
...
+  public class CapacityBoundedEventQueue extends ArrayBlockingQueue<Sequenced<? extends ChangeEvent>> {
+
+    private final long capacityLimit;
+    private long capacity;
+    /** Main lock guarding all access */
+    private final ReentrantLock lock;
+    /** Condition for waiting puts */
+    private final Condition notFull;
+
+    public CapacityBoundedEventQueue(int queueSize, long capcityLimit) {
```
albertshau: typo: capcity -> capacity

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
```diff
@@ -0,0 +1,179 @@
...
+      super(queueSize);
+      this.capacityLimit = capcityLimit;
+      this.capacity = 0;
+      this.lock = new ReentrantLock();
+      this.notFull = lock.newCondition();
+    }
+
+    private static void checkNotNull(Sequenced<? extends ChangeEvent> event) {
```
albertshau: how come this is needed, we shouldn't ever be adding null events right?

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
```diff
@@ -0,0 +1,179 @@
...
+      if (changeEvent instanceof DDLEvent) {
+        return 0;
+      }
+      return ((DMLEvent) changeEvent).getSizeInBytes();
+    }
+
+    @Override
+    public void put(Sequenced<? extends ChangeEvent> event) throws InterruptedException {
+      checkNotNull(event);
+      final ReentrantLock lock = this.lock;
```
albertshau: why do you need to do create this final variable instead of just referring to it directly?

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * A capacity bounded event queue. It will block if the capacity of the events in the queue reaches certain
+   * limit. When calculating the capacity of an event only the size of the row and previous row in DML event are
+   * calculated. Because other fields are fixed size and could be limited by queue size (the number of event in the
+   * queue)
+   */
+  public class CapacityBoundedEventQueue extends ArrayBlockingQueue<Sequenced<? extends ChangeEvent>> {
+
+    private final long capacityLimit;
+    private long capacity;
```
albertshau: nit: capacity is usually a number than decreases as usage increases. This number is being used as a counter for current size, so a name like 'usage' or 'size' would be more accurate.

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
```diff
@@ -0,0 +1,179 @@
...
+      lock.lockInterruptibly();
+      try {
+        int eventCapacity = computeCapacity(event);
+        while (capacity + eventCapacity > capacityLimit) {
+          if (nanos <= 0) {
+            return false;
+          }
+          nanos = notFull.awaitNanos(nanos);
+        }
+        boolean result = super.offer(event, nanos, TimeUnit.NANOSECONDS);
```
albertshau: to honor the contract more accurately, should actually be timing the loop above and subtracting that from the timeout sent here.

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    @Override
+    public void put(Sequenced<? extends ChangeEvent> event) throws InterruptedException {
+      checkNotNull(event);
+      final ReentrantLock lock = this.lock;
+      lock.lockInterruptibly();
+      try {
+        int eventCapacity = computeCapacity(event);
+        while (capacity + eventCapacity > capacityLimit) {
```
albertshau: if a single event is larger than the capacity limit, this will wait forever. Seems like it would be better to always try when the queue is empty. 

Immediately throwing an exception would also be appropriate if we just consider this class, but it wouldn't be nice for the replicator, since it would get into a continuous retry loop.

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
seanzhougoogle: good point:) done :)

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
seanzhougoogle: done :)

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
seanzhougoogle: Initially I added this just for completeness. I agree it's not useful in our use case, removed.

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
seanzhougoogle: agree, removed.

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
seanzhougoogle: make sense, thanks for the suggestion. Done :)

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
seanzhougoogle: it's already subtracted. The line 
`nanos = notFull.awitNanos(nanos)`
actually update the nanos by subtracting the waiting time.

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
seanzhougoogle: good point :)  done :)

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
albertshau: ah, got it

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -78,6 +80,14 @@ public String getRowId() {
     return rowId;
   }
 
+  public int getSizeInBytes() {
+    if (sizeInBytes < 0) {
+      // the size of all other fields are fixed, only return the dynamic size of row and previousRow
+     sizeInBytes = operation.getSizeInBytes() + computeSizeInBytes(previousRow);
```
albertshau: style: fix indentation to 2 spaces

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+
+  /**
+   * A capacity bounded event queue. It will block if the capacity of the events in the queue reaches certain
+   * limit. When calculating the capacity of an event only the size of the row and previous row in DML event are
+   * calculated. Because other fields are fixed size and could be limited by queue size (the number of event in the
+   * queue)
+   */
+  public class CapacityBoundedEventQueue extends ArrayBlockingQueue<Sequenced<? extends ChangeEvent>> {
```
albertshau: style: fix indendation (everything looks shifted by 2 extra spaces)

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
albertshau: this is still here, did you push all the updates?

albertshau (approved):
lgtm pending style fixes

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
seanzhougoogle: good catch, I missed some. now removed

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
seanzhougoogle: good catch, done :)

delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
seanzhougoogle: good catch, done :)

PR #124, opened 2021-02-02
CDAP-17373 pass tables to be replicated to assessor

seanzhougoogle:
What:

Pass information of the list of tables to be replicated to assessor

Why:
some delta source plugin need to know the list of tables when assessing the whole replicator pipeline

### DIFF
--- a/delta-api/src/main/java/io/cdap/delta/api/assessment/TableAssessorSupplier.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/assessment/TableAssessorSupplier.java
@@ -17,6 +17,9 @@
 package io.cdap.delta.api.assessment;
 
 import io.cdap.delta.api.Configurer;
+import io.cdap.delta.api.SourceTable;
+
+import java.util.List;
 
 /**
  * Creates TableAssessors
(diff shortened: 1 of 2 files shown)


delta-api/src/main/java/io/cdap/delta/api/assessment/TableAssessorSupplier.java
```diff
@@ -33,4 +36,17 @@
...
+
+  /**
+   * Create a table assessor that will check if there will be potential problems replicating tables.
+   *
+   * @param configurer configurer used to instantiate plugins
+   * @param tables tables to be replicated
+   * @return a table assessor
+   * @throws Exception if the table assessor could not be created
+   */
+  default TableAssessor<T> createTableAssessor(Configurer configurer, List<SourceTable> tables) throws Exception {
```
sagarkapare, line 48: This method returns `TableAssessor` which is meant for the single table only, right? How will the list of tables be used by `TableAssessor`? Can you provide some sample pseudo code for this API for my understanding?

seanzhougoogle: This is not for a single table , the TableAssessor has two interfaces :
 TableAssessment assess(T tableDescriptor).  -- assess a single table
Assessment assess() -- assess the whole replicator pipeline

sagarkapare (approved):

PR #125, opened 2021-02-18
CDAP-17491 Error out if the macro is used in the pipeline configuration as only secure store macro functions are supported.

sagarkapare:
JIRA: https://cdap.atlassian.net/browse/CDAP-17491

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/store/DefaultMacroEvaluator.java
+++ b/delta-app/src/main/java/io/cdap/delta/store/DefaultMacroEvaluator.java
@@ -43,11 +43,10 @@ public class DefaultMacroEvaluator implements MacroEvaluator {
   @Override
   @Nullable
   public String lookup(String property) {
-    String val = runtimeArgs.get(property);
-    if (val == null) {
-      throw new InvalidMacroException(String.format("Argument '%s' is not defined.", property));
-    }
-    return val;
+    String msg = String.format("Macro cannot be resolved for property '%s' as macro resolution is not supported " +
+                                 "for replication pipelines. Only secure store macro functions are supported.",
+                               property);
+    throw new InvalidMacroException(msg);
   }
 
   @Override
--- a/delta-app/src/main/java/io/cdap/delta/store/SystemServicePropertyEvaluator.java
+++ b/delta-app/src/main/java/io/cdap/delta/store/SystemServicePropertyEvaluator.java
@@ -26,9 +26,7 @@ import java.util.Map;
  * Evaluates macros in a map of properties.
  */
 public class SystemServicePropertyEvaluator implements PropertyEvaluator {
-  // only evaluate secure lookup macros
   private static final MacroParserOptions OPTIONS = MacroParserOptions.builder()
-    .disableLookups()
     .setFunctionWhitelist(DefaultMacroEvaluator.SECURE_FUNCTION_NAME)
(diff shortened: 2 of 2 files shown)


sagarkapare:
Error message looks like this after the fix - 
image

delta-app/src/main/java/io/cdap/delta/store/DefaultMacroEvaluator.java
```diff
@@ -43,11 +43,10 @@ public DefaultMacroEvaluator(Map<String, String> runtimeArgs, SecureStore secure
   @Override
   @Nullable
   public String lookup(String property) {
-    String val = runtimeArgs.get(property);
-    if (val == null) {
-      throw new InvalidMacroException(String.format("Argument '%s' is not defined.", property));
-    }
-    return val;
+    String msg = String.format("Macro cannot be resolved for property '%s' as macro resolution is not supported " +
```
seanzhougoogle, line 46: DefaultMacroEvaluator is also used below :

https://github.com/data-integrations/delta/blob/7ef527932d7ef4a9f830238a9d3e99916dd7e565/delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java#L182

I guess we only disable the use of Marco at Replicator design time (creating the replicator draft) but not at runtime (when the replicator is actually running) right ?

sagarkapare: I think its better to do it at runtime too for consistency. So if someone deploys pipeline bypassing the UI, it would result in the same error.

delta-app/src/main/java/io/cdap/delta/store/DefaultMacroEvaluator.java
```diff
@@ -43,11 +43,10 @@ public DefaultMacroEvaluator(Map<String, String> runtimeArgs, SecureStore secure
...
+                                 "for replication pipelines. Only secure store macro functions are supported.",
+                               property);
+    throw new RuntimeException(msg);
```
albertshau: better to throw an InvalidMacroException. Should avoid throwing Exception or RuntimeException in most circumstances.

albertshau: better to throw an InvalidMacroException. Should avoid throwing Exception or RuntimeException in most circumstances.

delta-app/src/main/java/io/cdap/delta/store/SystemServicePropertyEvaluator.java
```diff
@@ -28,7 +28,6 @@
 public class SystemServicePropertyEvaluator implements PropertyEvaluator {
   // only evaluate secure lookup macros
```
albertshau: this comment is no longer true

sagarkapare: fixed.

delta-app/src/main/java/io/cdap/delta/store/DefaultMacroEvaluator.java
sagarkapare: fixed.

albertshau (approved):

sagarkapare:
Green build here - https://builds.cask.co/browse/DEL-BT119-2

PR #126, opened 2021-02-22
Override the update supported method from Application.

sagarkapare:
Overriding the method `isUpdateSupported` in replication application to return `true`.

### DIFF
--- a/delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
+++ b/delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
@@ -226,8 +226,9 @@ public class DeltaConfig extends Config {
     for (Stage stage : getStages()) {
       upgradedStages.add(stage.updateStage(upgradeContext));
     }
-    return new DeltaConfig(description, upgradedStages, connections, resources, offsetBasePath,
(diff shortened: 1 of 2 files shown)


delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
```diff
@@ -223,11 +223,12 @@ public DeltaConfig updateConfig(ApplicationUpdateContext upgradeContext)
     throws Exception {
     List<Stage> upgradedStages = new ArrayList<>();
     // Upgrade all stages.
-    for (Stage stage : getStages()) {
+    for (Stage stage : this.getStages()) {
```
seanzhougoogle: nit: why do we need to add all those "this" ?

seanzhougoogle (approved):

delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
sagarkapare: removed `this`.

sagarkapare:
Build is green here - https://builds.cask.co/browse/DEL-BT120-2. Merging.

PR #134, opened 2021-04-29
resolve the event queue stucking issue

seanzhougoogle:
we have two limit of the queue, one is the capacity indicating the total bytes of the events in the queue and the other is the queue size indicating the total number of the events in the queue

the root cause of the issue is the line below 
https://github.com/data-integrations/delta/blob/ecd63ebb373651e1fabc02ac6081d61cb6b59f08/delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java#L75

initially `notFull` condition was only used to indicate the condition that when the capacity is reached. While above line can happen also when the # of events are reached limit. 

While the corresponding notify method like below :
https://github.com/data-integrations/delta/blob/ecd63ebb373651e1fabc02ac6081d61cb6b59f08/delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java#L130
is guarded by `if (eventSize > 0)`, thus if the event queue is filled with all those events that have size 0, when it's full, the producer thread will stuck forever, because the consumer thread will not notify the producer after it's consuming event from the queue.

and this will happen for delete events which we calculate their size as 0 (for metrics).

so the fix is to correctly calculate the size of delete event size and since we will never have 0 size event, we can use `notFull` to indicate the condition when either capacity or # of events reaches the limit.

### DIFF
--- a/delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
@@ -84,6 +84,10 @@ public class DMLEvent extends ChangeEvent {
     if (sizeInBytes < 0) {
       // the size of all other fields are fixed, only return the dynamic size of row and previousRow
       sizeInBytes = operation.getSizeInBytes() + computeSizeInBytes(previousRow);
+      // delete event didn't count the size of row in the DMLOperation
+      if (operation.getType() == DMLOperation.Type.DELETE) {
+        sizeInBytes += computeSizeInBytes(row);
+      }
     }
     return sizeInBytes;
   }
--- a/delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
+++ b/delta-app/src/main/java/io/cdap/delta/app/CapacityBoundedEventQueue.java
@@ -50,6 +50,8 @@ public class CapacityBoundedEventQueue extends ArrayBlockingQueue<Sequenced<? ex
     this.capacity = capacity;
     this.usage = 0;
     this.lock = new ReentrantLock();
+    // notFull indicates the condition when the event queue is full due to either the # of the events or the total size
+    // of the events
     this.notFull = lock.newCondition();
   }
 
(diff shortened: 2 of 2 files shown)


delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -84,6 +84,10 @@ public int getSizeInBytes() {
     if (sizeInBytes < 0) {
       // the size of all other fields are fixed, only return the dynamic size of row and previousRow
       sizeInBytes = operation.getSizeInBytes() + computeSizeInBytes(previousRow);
+      // delete event didn't count the size of row in the DMLOperation
+      if (operation.getType() == DMLOperation.Type.DELETE) {
```
sagarkapare, line 88: This will be used while emitting metrics too where we dont want to count row size for delete events.

seanzhougoogle: DMLOperation.getSizeInBytes is used for metrics :
https://github.com/data-integrations/delta/blob/develop/delta-app/src/main/java/io/cdap/delta/app/EventMetrics.java#L48

not this one :)

seanzhougoogle: and this one also calculate the size of previousRow which metrics don't care.

sagarkapare (approved):
PR lgtm 👍  One suggestion would be to run the pipeline with large amount of data similar to perf tests settings to make sure pipeline does not get stuck in between and process all records.

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
sagarkapare: ah ok. thanks Sean for clarification!

seanzhougoogle:
> PR lgtm 👍 One suggestion would be to run the pipeline with large amount of data similar to perf tests settings to make sure pipeline does not get stuck in between and process all records.

yeah, actually this bug was found during performance testing and I've verified the fix.

PR #135, opened 2021-05-03
CDAP-17917 add a delta reader interface for passing stop reason to stop method

seanzhougoogle:


delta-api/src/main/java/io/cdap/delta/api/EventReader.java
```diff
@@ -32,6 +32,21 @@
...
   default void stop() throws InterruptedException {
+    stop(StopReason.UNKNOWN);
+  }
+
+  /**
+   * Stop reading events.
+   * @param reason the reason why stop is requested.
+   * @throws InterruptedException if the reader was interrupted while stopping
+   */
+  default void stop(StopReason reason) throws InterruptedException {
```
sagarkapare: Would be good to have wrapper class `StopContext`  so that future additions can be done.
```
interface StopContext {
  
   public enum Origin {
      UNKNOWN,
      ERROR, // stop initiated because error
      USER // stop initiated by user
   }

   Origin getOrigin();
}
```

Then stop method could be 
```
default void stop(StopContext context) throws InterruptedException {}...
```

seanzhougoogle: thanks for the suggestion:) done :)

delta-api/src/main/java/io/cdap/delta/api/StopContext.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * The context of a stop request to the event reader
+ */
+public interface StopContext {
+  /**
+   * The origin of a stop request to the event reader
+   */
+  enum Origin {
+    UNKNOWN, ERROR, // stop initiated because error
```
sagarkapare: minor nit: move ERROR to new line.

sagarkapare (approved):

delta-api/src/main/java/io/cdap/delta/api/StopContext.java
seanzhougoogle: done :)

PR #136, opened 2021-05-03
CDAP-17586 don't assign sequence number to DDL event

seanzhougoogle:
sequence number is used to record the latest DML event applied in the target database. assigning sequence number to DDL event will cause some problem in some cases , see https://cdap.atlassian.net/jira/software/c/projects/CDAP/issues/CDAP-17586 for details.

### DIFF
--- a/delta-api/src/main/java/io/cdap/delta/api/Sequenced.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/Sequenced.java
@@ -26,18 +26,26 @@ import java.util.Objects;
 public class Sequenced<T> {
   private final T event;
   private final long sequenceNumber;
+  private static final long UNSUPPORTED_SEQUENCE = -1;
 
+  public Sequenced(T event) {
+    this(event, UNSUPPORTED_SEQUENCE);
+  }
   public Sequenced(T event, long sequenceNumber) {
     this.event = event;
     this.sequenceNumber = sequenceNumber;
   }
 
-  public T getEvent() {
-    return event;
+  public long getSequenceNumber() {
+    if (sequenceNumber == UNSUPPORTED_SEQUENCE) {
+      throw new UnsupportedOperationException();
+    }
+    return sequenceNumber;
   }
 
-  public long getSequenceNumber() {
-    return sequenceNumber;
+
+  public T getEvent() {
+    return event;
   }
 
   @Override
@@ -57,4 +65,5 @@ public class Sequenced<T> {
   public int hashCode() {
     return Objects.hash(event, sequenceNumber);
   }
+
 }
--- a/delta-api/src/main/java/io/cdap/delta/api/Sequenced.java
+++ b/delta-api/src/main/java/io/cdap/delta/api/Sequenced.java
@@ -28,6 +28,11 @@ public class Sequenced<T> {
   private final long sequenceNumber;
   private static final long UNSUPPORTED_SEQUENCE = -1;
 
+  /**
+   * Those events that don't support sequence number should use this constructor.
+   * e.g. DDL event
+   * @param event the event
+   */
   public Sequenced(T event) {
     this(event, UNSUPPORTED_SEQUENCE);
   }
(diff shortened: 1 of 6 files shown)


delta-app/src/main/java/io/cdap/delta/app/QueueingEventEmitter.java
```diff
@@ -64,7 +64,8 @@ public void emit(DDLEvent event) {
     }
 
     try {
-      eventQueue.put(new Sequenced<>(event, ++sequenceNumber));
+      // don't assign sequence number to DDL event
+      eventQueue.put(new Sequenced<>(event, -1));
```
sagarkapare: Assigning `-1` here seems error prone. It still means that target can access sequence number for DDL events using API `Sequence.getSequenceNumber()` and accidentally use it. We need to make sure that target is aware of the contract `DDL Events wont have sequence number`. One possible way may be `Sequenced` class can be modified as :
```java
public class Sequenced {
   private final int UNSUPPORTED_SEQUENCE = -1;
  public Sequenced(T event) {
    this(event, UNSUPPORTED_SEQUENCE);
  } 
   public Sequenced(T event, long sequenceNumber) {
    this.event = event;
    this.sequenceNumber = sequenceNumber;
  }
public long getSequenceNumber() {
    if (sequenceNumber == UNSUPPORTED_SEQUENCE) {
        throw new UnsupportedOperationExcetpion();
    }
    return sequenceNumber;
  }
}
```

then in here we can do `evenrQueue.put(new Sequenced<>(event))`

seanzhougoogle: good suggestion :). Done :)

delta-api/src/main/java/io/cdap/delta/api/Sequenced.java
```diff
@@ -26,20 +26,28 @@
...
   public Sequenced(T event, long sequenceNumber) {
     this.event = event;
     this.sequenceNumber = sequenceNumber;
   }
 
-  public T getEvent() {
-    return event;
-  }
-
   public long getSequenceNumber() {
```
sagarkapare, line 48: add javadoc about when it will throw an exceptipn?

delta-api/src/main/java/io/cdap/delta/api/Sequenced.java
```diff
@@ -26,20 +26,28 @@
 public class Sequenced<T> {
   private final T event;
   private final long sequenceNumber;
+  private static final long UNSUPPORTED_SEQUENCE = -1;
 
+  public Sequenced(T event) {
```
sagarkapare, line 36: javadoc would be good indicating when to use this constructor vs next.

delta-api/src/main/java/io/cdap/delta/api/Sequenced.java
```diff
@@ -26,20 +26,28 @@
...
     this.sequenceNumber = sequenceNumber;
   }
 
-  public T getEvent() {
-    return event;
-  }
-
   public long getSequenceNumber() {
+    if (sequenceNumber == UNSUPPORTED_SEQUENCE) {
+      throw new UnsupportedOperationException();
```
sagarkapare: would be good to add message for the exception.

sagarkapare (approved):
minor comments about Javadoc. else LGTM

delta-api/src/main/java/io/cdap/delta/api/Sequenced.java
seanzhougoogle: done :)

delta-api/src/main/java/io/cdap/delta/api/Sequenced.java
seanzhougoogle: done :)

delta-api/src/main/java/io/cdap/delta/api/Sequenced.java
seanzhougoogle: done :)

PR #147, opened 2021-06-23
Expose the number of worker instances to source and targets.

sagarkapare:
For Oracle golden gate source, there has to be 1:1 mapping between worker instances and xstream outbound servers. One instance cannot talk to more than one outbound server, similarly outbound server cannot accept more than one connection. 
Exposing number of instances the worker is configured with to the source plugins so that source can fail fast if there is mismatch.


seanzhougoogle (approved):

sagarkapare:
https://builds.cask.co/browse/DEL-BT142 Build is green. Merging.

PR #144, opened 2021-06-11
CDAP-18057 Skip the latency checks for now.

sagarkapare:
Disabling latency checks. JIRA for debugging the failure - https://cdap.atlassian.net/browse/CDAP-18057

PR #149, opened 2021-07-20
CDAP-18057 Add aritificial delay to ingest time in test case so that non-zero latency can be verified.

sagarkapare:
Reverting changes from PR #144


seanzhougoogle (approved):

sagarkapare:
Build is green - https://builds.cask.co/browse/DEL-BT144-1 merging.

PR #154, opened 2021-10-01
issue/CDAP-18314 : Support for Db(SQL) based StateStore

sahusanket:


delta-app/pom.xml
```diff
@@ -47,6 +47,12 @@
       <version>${cdap.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-proto</artifactId>
+      <version>${cdap.version}</version>
+      <scope>provided</scope>
```
albertshau: is this being used anywhere other than tests? If so, which class? If not, this should be test scope.

delta-proto/src/main/java/io/cdap/delta/proto/StateStoreConfig.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.Objects;
+
+/**
+ * The class representing the config related to initialization of state store.
+ */
+public class StateStoreConfig {
+  private final String type;
+
+  public StateStoreConfig(String type) {
```
seanzhougoogle: better to make the type an enum ?

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
```diff
@@ -0,0 +1,154 @@
...
+      return new String(ByteStreams.toByteArray(is), StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      return "Unknown error due to failure to read from error output: " + e.getMessage();
+    }
+  }
+
+  /**
+   * A connection object is created with the given System Service paths and parameters
+   *
+   * @param methodPath
```
seanzhougoogle: private methods do not necessary to have javadoc, but if you want to add, it would be better to add some doc for the params listed here :)

delta-app/src/main/java/io/cdap/delta/app/PipelineStateService.java
```diff
@@ -57,6 +57,10 @@ public PipelineStateService(DeltaWorkerId id, StateStore stateStore) {
    */
   public void load() throws IOException {
     byte[] bytes = stateStore.readState(id, STATE_KEY);
+    load(bytes);
+  }
+
+  public void load(byte[] bytes) {
```
albertshau: this should not be exposed, it breaks encapsulation. This class is in charge of how the data is represented, exposing this makes it so that the caller also needs to know.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -74,24 +78,28 @@
...
   protected void configure() {
     setProperties(Collections.singletonMap(OFFSET_PATH, offsetBasePath));
+    setProperties(Collections.singletonMap(STATE_STORE_CONFIG, GSON.toJson(stateStoreConfig)));
   }
 
   @GET
   @Path("v1/contexts/{context}/drafts")
   public void listDrafts(HttpServiceRequest request, HttpServiceResponder responder,
                          @PathParam("context") String namespaceName) {
-    respond(namespaceName, responder, (draftService, namespace) -> {
```
albertshau, line 93: why make this change?

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -245,25 +254,45 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
-      StateStore stateStore = StateStore.from(path);
+      StateStoreConfig stateStoreConfig = GSON.fromJson(
+        getContext().getSpecification().getProperty(STATE_STORE_CONFIG), StateStoreConfig.class);
 
-      Long latestGen = stateStore.getLatestGeneration(namespaceName, stateRequest.getName());
-      // this can happen if the pipeline was never started
-      if (latestGen == null) {
-        responder.sendString(GSON.toJson(PipelineReplicationState.EMPTY));
-        return;
+      if (stateStoreConfig.getType().equals(StateStoreFactory.DB_KEY)) {
```
albertshau: the handler should not be doing this type logic. It should not need to understand what different state store implementations exist, it should just get it using the factory.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -276,7 +305,15 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
       for (int instanceId : workerInstances) {
         PipelineStateService stateService = new PipelineStateService(new DeltaWorkerId(pipelineId, instanceId),
                                                                      stateStore);
-        stateService.load();
+        if (stateStoreConfig.getType().equals(StateStoreFactory.DB_KEY)) {
```
albertshau: should not do this. There should not be any logic that checks the state store type except in the factory, where it needs to determine which underlying implementation to use.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +325,99 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+    Methods to manage offset and other states
+   */
+
+  /**
+   * Returns the State Store Type to be used by the workers to store offset and state
+   */
+  @GET
+  @Path("v1/statestoretype")
+  public void getStateStoreType(HttpServiceRequest request, HttpServiceResponder responder) {
+      responder.sendString(getContext().getSpecification().getProperty(STATE_STORE_CONFIG));
```
albertshau: indentation should be 2 spaces

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +325,99 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
 
+  /**
+    Methods to manage offset and other states
+   */
+
+  /**
+   * Returns the State Store Type to be used by the workers to store offset and state
+   */
+  @GET
+  @Path("v1/statestoretype")
```
albertshau: statestoretype -> statestore. this is not the type, it's the entire state store information

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +325,99 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+  public void getStateStoreType(HttpServiceRequest request, HttpServiceResponder responder) {
+      responder.sendString(getContext().getSpecification().getProperty(STATE_STORE_CONFIG));
+  }
+
+  /**
+   * Looks up the delta_offset_store table for offset data.
+   */
+  @GET
+  @Path("v1/context/{namespace}/apps/{app}/generations/{generation}/instanceIds/{instanceid}/offset")
```
albertshau: contexts

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +325,99 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+    respond(namespaceName, responder, (namespace) -> {
+      responder.sendString(getSQLStoreService().readOffsetData(namespaceName, app, generation, instanceid));
+    });
+  }
+
+  /**
+   * Upserts in the delta_offset_store table for offset data.
+   */
+  @PUT
+  @Path("v1/context/{namespace}/apps/{appname}/generations/{generation}/instanceIds/{instanceid}/offset")
```
albertshau: contexts

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +325,99 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+   */
+  @GET
+  @Path("v1/context/{namespace}/apps/{app}/generations/{generation}/instanceIds/{instanceid}/offset")
+  public void getOffsetData(HttpServiceRequest request, HttpServiceResponder responder,
+                           @PathParam("namespace") String namespaceName,
+                           @PathParam("app") String app,
+                           @PathParam("generation") long generation,
+                           @PathParam("instanceid") int instanceid) {
+    respond(namespaceName, responder, (namespace) -> {
+      responder.sendString(getSQLStoreService().readOffsetData(namespaceName, app, generation, instanceid));
```
albertshau: similar comment here and everywhere below, handler should never know about the state store implementation

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
```diff
@@ -0,0 +1,154 @@
...
+  private static final double RETRY_DELAY_MULTIPLIER = 1.2d;
+  private static final double RETRY_RANDOMIZE_FACTOR = 0.1d;
+
+  ServiceDiscoverer serviceDiscoverer;
+
+  public RemoteHttpClient(ServiceDiscoverer context) {
+    this.serviceDiscoverer = context;
+  }
+
+  public String retryableApiCall(String connectionUrl, String ...method) {
```
seanzhougoogle: I think we already introduced FailSafe  in the project and you don't need to reinvent the wheel :
https://github.com/data-integrations/delta/blob/4cd6ca136124eb641f40c4bc94055a519f56b93f/delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java#L249

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentService.java
```diff
@@ -17,23 +17,35 @@
...
     this.offsetBasePath = offsetBasePath;
+    this.stateStoreConfig = stateStoreConfig;
   }
 
   @Override
   protected void configure() {
     setName(NAME);
-    addHandler(new AssessmentHandler(offsetBasePath));
+    // If the value is not populated in cdc capability json, then fall back to old state-store[HCFS].
+    if (stateStoreConfig == null) {
```
albertshau: this logic shouldn't be here. The DeltaConfig getter should be the one checking for null and returning the proper default value.

delta-app/src/main/java/io/cdap/delta/store/DBReplicationOffsetStore.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Optional;
+
+/**
+ * A StructuredTable based database table to store replication offset and sequence data.
+ */
+public class DBReplicationOffsetStore {
+
+  private static final StructuredTableId TABLE_ID = new StructuredTableId("delta_offset_store");
+  private static final String NAMESPACE_COL = "namespace";
+  private static final String GENERATION_COL = "generation";
```
albertshau: generation is used in other places to mean the namespace generation. Would be good to name this app_generation

albertshau: Also, should we be including the namespace generation as a column? In other words, if somebody deletes a namespace, should all the offsets for the namespace be deleted? Seems like it should, but haven't thought through it much

delta-app/src/main/java/io/cdap/delta/store/DBReplicationOffsetStore.java
```diff
@@ -0,0 +1,103 @@
...
+    try {
+      StructuredTable table = context.getTable(TABLE_ID);
+      return new DBReplicationOffsetStore(table);
+    } catch (TableNotFoundException e) {
+      throw new IllegalStateException(String.format(
+        "System table '%s' does not exist. Please check your system environment.", TABLE_ID.getName()), e);
+    }
+  }
+
+  public String getOffsets(String namespace, String appName, Long generation, int instanceId)
```
albertshau: annotate method as `@Nullable`

delta-app/src/main/java/io/cdap/delta/store/DBStateStoreController.java
```diff
@@ -0,0 +1,88 @@
...
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import javax.annotation.Nullable;
+import javax.ws.rs.HttpMethod;
+
+/**
+ * Class to handle Rest Api calls to System Service [ Assessor ]
+ */
+
+public class DBStateStoreController implements StateStore {
```
albertshau: naming this something like RemoteStateStore would be more accurate. 

The actual implementation is not known to this class. It could technically still be HCFS backed.

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -172,8 +180,8 @@ public void initialize(WorkerContext context) throws Exception {
     DeltaWorkerId id = new DeltaWorkerId(new DeltaPipelineId(context.getNamespace(), appSpec.getName(), generation),
                                          context.getInstanceId());
 
-    Path path = new Path(offsetBasePath);
-    StateStore stateStore = StateStore.from(path);
+    StateStore stateStore = new StateStoreFactory()
+      .getInstance(stateStoreConfig.getType(), context, offsetBasePath);
```
seanzhougoogle: should 'getInstance' be a static method ?

delta-app/src/main/java/io/cdap/delta/store/DBStateStoreService.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public String readStateData(String namespace, String appName, Long generation, int instanceId, String state) {
+    return TransactionRunners.run(txRunner, context -> {
+      return DBReplicationStateStore.get(context).getStateData(namespace, appName, generation, instanceId, state);
+    });
+  }
+
+  public void saveStateData(String namespace, String appName, Long generation, int instanceId, String key,
+                            String data) {
+     TransactionRunners.run(txRunner, context -> {
```
albertshau: indentation

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
```diff
@@ -0,0 +1,140 @@
...
+    try {
+      StructuredTable table = context.getTable(TABLE_ID);
+      return new DBReplicationStateStore(table);
+    } catch (TableNotFoundException e) {
+      throw new IllegalStateException(String.format(
+        "System table '%s' does not exist. Please check your system environment.", TABLE_ID.getName()), e);
+    }
+  }
+
+  public String getStateData(String namespace, String appName, Long generation, int instanceId, String state)
```
albertshau: shouldn't this be byte[]?

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
```diff
@@ -0,0 +1,140 @@
...
+    List<Field<?>> keys = getKey(namespace, appName, generation, instanceId, state);
+    Optional<StructuredRow> row = table.read(keys);
+    if (!row.isPresent() || row.get().getString(VALUE_COL) == null) {
+      return StringUtils.EMPTY;
+    }
+    return row.get().getString(VALUE_COL);
+  }
+
+  public void writeStateData(String namespace, String appName, Long generation, int instanceId, String state,
+                             String data) throws IOException {
```
albertshau: shouldn't this be byte[]?

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
```diff
@@ -0,0 +1,140 @@
...
+  private static final String UPDATED_COL = "last_updated";
+
+  public static final StructuredTableSpecification TABLE_SPEC = new StructuredTableSpecification.Builder()
+    .withId(TABLE_ID)
+    .withFields(new FieldType(NAMESPACE_COL, FieldType.Type.STRING),
+                new FieldType(GENERATION_COL, FieldType.Type.LONG),
+                new FieldType(APPNAME_COL, FieldType.Type.STRING),
+                new FieldType(INSTANCEID_COL, FieldType.Type.INTEGER),
+                new FieldType(STATE_COL, FieldType.Type.STRING),
+                new FieldType(VALUE_COL, FieldType.Type.STRING),
```
albertshau: shouldn't this be bytes?

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
```diff
@@ -0,0 +1,140 @@
...
+  private static final String APPNAME_COL = "app_name";
+  private static final String INSTANCEID_COL = "instance_id";
+  private static final String STATE_COL = "state";
+  private static final String VALUE_COL = "value";
+  private static final String UPDATED_COL = "last_updated";
+
+  public static final StructuredTableSpecification TABLE_SPEC = new StructuredTableSpecification.Builder()
+    .withId(TABLE_ID)
+    .withFields(new FieldType(NAMESPACE_COL, FieldType.Type.STRING),
+                new FieldType(GENERATION_COL, FieldType.Type.LONG),
```
albertshau: similar comment/question about namespace generation

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
```diff
@@ -0,0 +1,140 @@
...
+  public void writeStateData(String namespace, String appName, Long generation, int instanceId, String state,
+                             String data) throws IOException {
+    Collection<Field<?>> fields = getKey(namespace, appName, generation, instanceId, state);
+    fields.add(Fields.stringField(VALUE_COL, data));
+    fields.add(Fields.longField(UPDATED_COL, System.currentTimeMillis()));
+
+    table.upsert(fields);
+  }
+
+  public Long getLatestGeneration(String namespace, String appName) throws IOException {
```
albertshau, line 100: why is this needed?

delta-app/src/main/java/io/cdap/delta/store/DBStateStoreController.java
```diff
@@ -0,0 +1,88 @@
...
+  public void writeOffset(DeltaWorkerId id, OffsetAndSequence offset) {
+    String requestBody = GSON.toJson(offset);
+    remoteHttpClient.retryableApiCall(getMethodPath(id, OFFSET_KEY), HttpMethod.PUT, requestBody);
+  }
+
+  @Override
+  public byte[] readState(DeltaWorkerId id, String key) {
+    String byteString = remoteHttpClient.retryableApiCall(getMethodPathState(id, key), HttpMethod.GET);
+    byte[] byteArray = byteString.getBytes(StandardCharsets.UTF_8);
+    return byteArray;
```
albertshau: nit: can just return on the previous line

delta-app/src/main/java/io/cdap/delta/store/DBStateStoreController.java
```diff
@@ -0,0 +1,88 @@
...
+
+  @Override
+  public OffsetAndSequence readOffset(DeltaWorkerId id) {
+    String offsetAndSequenceJsonStr =
+      remoteHttpClient.retryableApiCall(getMethodPath(id, OFFSET_KEY), HttpMethod.GET);
+    if (offsetAndSequenceJsonStr.equals("")) {
+      return null;
+    }
+    OffsetAndSequence offsetAndSequence = GSON.fromJson(offsetAndSequenceJsonStr, OffsetAndSequence.class);
+    return offsetAndSequence;
```
albertshau: nit: can just return on previous line

delta-app/src/main/java/io/cdap/delta/store/HCFSStateStore.java
```diff
@@ -0,0 +1,151 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
```
albertshau: fix copyright year

delta-app/src/main/java/io/cdap/delta/store/HCFSStateStore.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Stores replicator state based on hadoop filesystem .
+ */
+public class HCFSStateStore implements StateStore {
```
albertshau, line 44: is this just the previous StateStore implementation except renamed? If not what are the differences?

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
```diff
@@ -0,0 +1,154 @@
...
+    double maxMultiplier = RETRY_DELAY_MULTIPLIER + RETRY_DELAY_MULTIPLIER * RETRY_RANDOMIZE_FACTOR;
+    Stopwatch stopWatch = new Stopwatch().start();
+
+    try {
+      while (stopWatch.elapsedTime(TimeUnit.MILLISECONDS) < TIMEOUT_MILLIS) {
+        try {
+          if (method[0].equals(HttpMethod.GET)) {
+            return executeGetCall(getHttpUrlConnection(connectionUrl));
+          } else if (method[0].equals(HttpMethod.POST) || method[0].equals(HttpMethod.PUT)) {
+            return executeCallWithBody(getHttpUrlConnection(connectionUrl), method[0], method[1]);
```
albertshau: don't pass arguments as an array with special meaning in each array slot.

Make the method a separate argument of type HttpMethod and the request body as a `@Nullable String`

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
```diff
@@ -0,0 +1,154 @@
...
+  private String executeCallWithBody(HttpURLConnection urlConn, String method, String body) throws IOException {
+
+    urlConn.setDoOutput(true);
+    urlConn.setRequestMethod(method);
+
+    OutputStream os = urlConn.getOutputStream();
+    OutputStreamWriter osw = new OutputStreamWriter(os, "UTF-8");
+    osw.write(body);
+    osw.flush();
+    osw.close();
```
albertshau: close() should usually be in a try-finally. An exception above will cause these not to be closed.

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
```diff
@@ -0,0 +1,154 @@
...
+    return validateAndRetrieveContent(urlConn);
+  }
+
+  /**
+   * Copied from "AbstractServiceRetryableMacroEvaluator"
+   * Returns the full content of the error stream for the given {@link HttpURLConnection}.
+   */
+  private String validateAndRetrieveContent(HttpURLConnection urlConn) throws IOException {
+    String serviceName = "State Store (AssessorService)";
+    if (urlConn == null) {
```
albertshau: any time an argument can be null, it should be annotated as `@Nullable`

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    int responseCode = urlConn.getResponseCode();
+    if (responseCode != HttpURLConnection.HTTP_OK) {
+      throw new RetryableException("Failed to call " + serviceName + " service with status " + responseCode + ": " +
+                                     getError(urlConn));
+    }
+
+    try (Reader reader = new InputStreamReader(urlConn.getInputStream(), StandardCharsets.UTF_8)) {
+      return CharStreams.toString(reader);
```
albertshau: returning a String and then parsing that string is not as efficient as decoding it into an object directly from the stream. See the Gson methods that take a Reader as input.

It would be better if the return type of this method is a generic, with the Type passed in as an argument.

delta-app/src/main/java/io/cdap/delta/store/StateStoreFactory.java
```diff
@@ -0,0 +1,40 @@
...
+ * This factory initiates the state store based on the new sql feature flag (which is environment dependent)
+ */
+public class StateStoreFactory {
+
+  public static final String DB_KEY = "SQL";
+  public static final String HCFS_KEY = "HCFS";
+
+  public StateStore getInstance(String stateStoreType, WorkerContext context, String offsetBasePath)
+    throws IOException {
+    if (stateStoreType != null && stateStoreType.toUpperCase(Locale.ROOT).trim().equals(DB_KEY)) {
```
albertshau: this is equivalent to `DB_KEY.equalsIgnoreCase(stateStoreType)`

albertshau: However, it's weird that somebody could put in a value like "asldfknnvz" and it would result in the HCFS store. 

It would be better to have an enum and pass that in here as the type. Validation of the type should occur in the app's configure() method.

delta-app/src/test/java/io/cdap/delta/app/DeltaPipelineDBStateStoreTest.java
```diff
@@ -0,0 +1,680 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Tests for delta pipelines
+ */
+public class DeltaPipelineDBStateStoreTest extends DeltaPipelineTestBase {
```
albertshau: Seems like these are copies of the existing tests? Should refactor common code to a common place, not copy-paste.

albertshau:
next time, please break up your PRs.

For example, one way to break this one up would have been to do one that introduces the StateStore interface and moves the HCFS implementation under it, then another to add the DB implementation, then another for hooking everything up together.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentService.java
```diff
@@ -17,23 +17,35 @@
...
+      stateStoreConfig = new StateStoreConfig(StateStoreFactory.HCFS_KEY);
+    }
+    addHandler(new AssessmentHandler(offsetBasePath, stateStoreConfig));
     createTable(DraftStore.TABLE_SPEC);
+    createTable(DBReplicationStateStore.TABLE_SPEC);
```
seanzhougoogle, line 36: check the stateStoreConfig before creating those tables. For HCFS store, you don't need to create those tables, right ?

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +325,99 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
     }));
   }
 
+  /**
```
seanzhougoogle: use a normal comment instead of javadoc comment

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
```diff
@@ -0,0 +1,140 @@
...
+    .withPrimaryKeys(NAMESPACE_COL, APPNAME_COL, GENERATION_COL, INSTANCEID_COL, STATE_COL)
+    .build();
+
+  private final StructuredTable table;
+
+  private DBReplicationStateStore(StructuredTable table) {
+    this.table = table;
+  }
+
+  static DBReplicationStateStore get(StructuredTableContext context) {
```
seanzhougoogle, line 72: why isn't this factory method public while the Class itself is public ?

delta-app/src/main/java/io/cdap/delta/store/DBStateStoreController.java
seanzhougoogle: rename it to `DBStateStore`, have a `Controller` suffix looks wired here as an implementation of StateStore

delta-app/src/main/java/io/cdap/delta/store/DBStateStoreController.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class DBStateStoreController implements StateStore {
+
+  private static final String BASE_PATH = "v1";
+  private static final String OFFSET_KEY = "offset";
+
+  private RemoteHttpClient remoteHttpClient;
+  private static final Gson GSON = new Gson();
+
+  public DBStateStoreController(ServiceDiscoverer context) {
```
seanzhougoogle: context -> `serviceDiscoverer`

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
```diff
@@ -0,0 +1,154 @@
...
+import java.io.Reader;
+import java.net.HttpURLConnection;
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.TimeUnit;
+import javax.ws.rs.HttpMethod;
+
+/**
+ * Class to handle Api calls to System Service
+ */
+public class RemoteHttpClient {
```
seanzhougoogle: this name looks too general while it's not a general httpclient, it's calling a specific service. Shall we call it AssessmentServiceClient ?

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +325,99 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+    Methods to manage offset and other states
```
seanzhougoogle: why are we putting them together with Assessor Service ? It looks like the newly added APIs are self-contained and could be put in a standalone service.

delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
seanzhougoogle: wrap all those statestore initialization logic in one method ?

seanzhougoogle (commented):
It's a huge PR. Went through them roughly today with some comments and will continue to review and add more comments tomorrow :)

delta-app/pom.xml
sahusanket: It is used in RemoteHttpClient  only to get the string "SYSTEM" from the static variable , while opening a connection. "NamespaceId.SYSTEM.getNamespace()" . 

did not want to hardcode it.

delta-app/src/main/java/io/cdap/delta/store/StateStoreFactory.java
sahusanket: The idea was that for non-ODF environment, even if the CDC json is not populated, It will default to HCFS statestore.
But yeah I will make it ENUM , but default "NULL"  to HCFS .

sahusanket: A workercontext is needed to access the cdc.json config since, it only available to App via REST calls. 
So, in configure method, there is no workercontext to access the system service. Hence in configure method, I donot have the value of the cdc.json config.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
sahusanket: Was trying to make it generic for both "DraftService" , "DBStateStoreService" , since the functionality is exactly same.

delta-app/src/main/java/io/cdap/delta/store/DBReplicationOffsetStore.java
sahusanket: Yes it will be deleted once we have the deletion hook. 

But it causes no harm to future pipelines, as the "app_generation" will be always incrementing and unique. there won't be a conflict of offset.

delta-app/src/main/java/io/cdap/delta/store/HCFSStateStore.java
sahusanket: It's just the rename. No other difference

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
sahusanket: It is because of the limited no of IPs left, we are not adding a new service for this.

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
sahusanket: The intention was to keep this `get` method usable with the package only. 

While the class is public because of the `StructuredTableSpecification TABLE_SPEC` is being used by `AssessmentService` , which is outside of the package.

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
sahusanket: This was kind of discussed in review meeting, 
the conclusion was that incase it's just a API failure, it should just recover from here it self, which will save time and effort, instead of restarting the worker process.

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
sahusanket: This was only for the Unit test actually.

delta-app/src/main/java/io/cdap/delta/store/DBStateStoreController.java
sahusanket: yes. 
I renamed to "RemoteStateStore" as suggested by albert previously.

sahusanket: As of now, 

REST end points in assessor service were only meant for DB level operations .. i.e. the new ones 

While the HCFS will used in the same way as before, without any change. 
The assessor service is only dealing with SQL type state store only. 

are you suggesting to shift HCFS operations via assessor service as well ?

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
seanzhougoogle: I was not talking about restarting the worker process. FailSafe is a library you can use for retry (including backoff) service calls instead of implementing it yourself here which is kind of reinventing the wheel. The above link I gave is just an example about how FaileSafe is used.

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
sahusanket: only `OffsetAndSequence` gson conversion is done within replication App, and it happens once on every start of worker. Other states deal with byte[]

Also tried the above approach the type `byte[].class` for other states is giving issues in conversion from reader.

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
sahusanket: Oh got it. 
The code looks much cleaner now. Thank you !

delta-app/pom.xml
albertshau: cdap-proto is actually an internal module in CDAP. It isn't exposed to application by the CDAP platform. Only modules with 'api' in the name are exposed.

I'm a bit surprised that the code runs. Is cdap-proto still being packaged in the delta-app jar somehow? That being said, if it's just being used for "system", I think it's ok to hardcode that value. Not worth packaging this dependency in the jar.

delta-app/src/main/java/io/cdap/delta/store/DBReplicationOffsetStore.java
albertshau: ah right, makes sense.

delta-app/src/main/java/io/cdap/delta/store/DBStateStoreController.java
albertshau: All this class knows is that it is implemented by making remote calls to the service. Whatever the service does is invisible to the class and should not be included in the naming.

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
albertshau: if its for unit test, annotate it with `@VisibleForTesting`. It should also usually be package private instead of public

delta-app/src/main/java/io/cdap/delta/store/RemoteHttpClient.java
albertshau: that's because bytes are not json.

If it's handling bytes, it definitely shouldn't be converting to a string. Have one method for the bytes and another for json.

delta-app/src/main/java/io/cdap/delta/store/StateStoreFactory.java
albertshau: I'm not sure I understand how the config is related to using an enum here. Can you describe what the problem is?

albertshau:
next time, please don't squash commits until reviews are completed. That way, we can review just the changes instead of the entire thing.

delta-app/src/main/java/io/cdap/delta/app/PipelineStateService.java
```diff
@@ -39,7 +39,7 @@
  */
 public class PipelineStateService {
   private static final Gson GSON = new Gson();
-  private static final String STATE_KEY = "pipeline";
+  public static final String STATE_KEY = "pipeline";
```
albertshau: don't see it being used anywhere, should keep it private

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
albertshau: They can go in a different Handler though, for cleaner separation of logic. Handlers all run within the same service container.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +285,88 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
     }));
   }
 
+  /*
+    Methods to manage offset and other states
+   */
+  //Looks up the delta_offset_store table for offset data.
+  @GET
+  @Path("v1/contexts/{namespace}/apps/{app}/generations/{generation}/instanceIds/{instanceid}/offset")
```
albertshau: instanceIds -> instances. Same for other paths

This is because in a REST path, it represents a collection of resources. The resource is an instance and not an id.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +285,88 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+  public void saveOffsetData(HttpServiceRequest request, HttpServiceResponder responder,
+                             @PathParam("namespace") String namespaceName,
+                             @PathParam("app") String app,
+                             @PathParam("generation") long generation,
+                             @PathParam("instanceid") int instanceid) {
+    respond(namespaceName, responder, (namespace) -> {
+      DeltaWorkerId id = new DeltaWorkerId(new DeltaPipelineId(namespaceName, app, generation), instanceid);
+      String body = StandardCharsets.UTF_8.decode(request.getContent()).toString();
+      getStateStore().writeOffset(id, GSON.fromJson(body, OffsetAndSequence.class));
```
albertshau: should be decoding it into an OffsetAndSequence using a Reader, not by converting first into a String and then into the object.

Handlers should assume the user can pass in malformed input and validate all of it before proceeding. Does respond() catch exceptions decoding the request body and return a 400? Should also do some validation to make sure it's a valid offset and sequence, for example by checking that the object itself is not null and any required fields are not null.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +285,88 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
   }
 
+  /*
+    Methods to manage offset and other states
+   */
+  //Looks up the delta_offset_store table for offset data.
+  @GET
+  @Path("v1/contexts/{namespace}/apps/{app}/generations/{generation}/instanceIds/{instanceid}/offset")
+  public void getOffsetData(HttpServiceRequest request, HttpServiceResponder responder,
+                           @PathParam("namespace") String namespaceName,
```
albertshau: indentation

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +285,88 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+      getStateStore().writeOffset(id, GSON.fromJson(body, OffsetAndSequence.class));
+      responder.sendStatus(HttpURLConnection.HTTP_OK);
+    });
+  }
+
+  //Looks up the delta_state_store table for the given state
+  @GET
+  @Path("v1/contexts/{namespace}/apps/{app}/generations/{generation}/instanceIds/{instanceid}/states/{state}")
+  public void getStateData(HttpServiceRequest request, HttpServiceResponder responder,
+                        @PathParam("namespace") String namespaceName,
```
albertshau: indentation

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +285,88 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+                        @PathParam("app") String app,
+                        @PathParam("generation") long generation,
+                        @PathParam("instanceid") int instanceid,
+                        @PathParam("state") String state) {
+    respond(namespaceName, responder, (namespace) -> {
+      DeltaWorkerId id = new DeltaWorkerId(new DeltaPipelineId(namespaceName, app, generation), instanceid);
+      byte[] data = getStateStore().readState(id, state);
+      responder.sendString(new String(data, StandardCharsets.UTF_8));
```
albertshau: don't send bytes as a string, it is incorrect to do so. There is a send method that takes a ByteBuffer, use that instead.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +285,88 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+    });
+  }
+
+  // Upserts in the delta_state_store table for state data.
+  @PUT
+  @Path("v1/contexts/{namespace}/apps/{app}/generations/{generation}/instanceIds/{instanceid}/states/{state}")
+  public void saveStateData(HttpServiceRequest request, HttpServiceResponder responder,
+                         @PathParam("namespace") String namespaceName,
```
albertshau: indentation

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +285,88 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+                         @PathParam("app") String app,
+                         @PathParam("generation") long generation,
+                         @PathParam("instanceid") int instanceid,
+                         @PathParam("state") String state) {
+
+    respond(namespaceName, responder, (namespace) -> {
+      DeltaWorkerId id = new DeltaWorkerId(new DeltaPipelineId(namespaceName, app, generation), instanceid);
+      ByteBuffer byteBuffer = request.getContent();
+      byte[] data = new byte[byteBuffer.remaining()];
+      byteBuffer.get(data);
```
albertshau: This way will not reset the ByteBuffer position. Use `Bytes.toBytes()` instead of manually doing this.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +285,88 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+      getStateStore().writeState(id, state, data);
+      responder.sendStatus(HttpURLConnection.HTTP_OK);
+    });
+  }
+
+  //Fetches the latest generation for the given namespace and app.
+  @GET
+  @Path("v1/context/{namespace}/apps/{app}/maxgeneration")
+  public void getMaxGeneration(HttpServiceRequest request, HttpServiceResponder responder,
+                           @PathParam("namespace") String namespaceName,
```
albertshau: indentation

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -288,6 +285,88 @@ public void getState(HttpServiceRequest request, HttpServiceResponder responder,
...
+      byteBuffer.get(data);
+
+      getStateStore().writeState(id, state, data);
+      responder.sendStatus(HttpURLConnection.HTTP_OK);
+    });
+  }
+
+  //Fetches the latest generation for the given namespace and app.
+  @GET
+  @Path("v1/context/{namespace}/apps/{app}/maxgeneration")
```
albertshau: this seems like it shouldn't be exposed. Where is it being used?

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -0,0 +1,145 @@
...
+import java.time.temporal.ChronoUnit;
+import javax.annotation.Nullable;
+import javax.ws.rs.HttpMethod;
+
+/**
+ * Class to handle Api calls to System Service
+ */
+public class AssessmentServiceClient {
+
+  ServiceDiscoverer serviceDiscoverer;
```
albertshau: private final

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -0,0 +1,145 @@
...
+
+  private String executeCallWithBody(HttpURLConnection urlConn, String method, String body)
+    throws IOException {
+
+    OutputStream os = null;
+    OutputStreamWriter osw = null;
+
+    try {
+      urlConn.setDoOutput(true);
+      urlConn.setRequestMethod(method);
```
albertshau: If either this or setDoOutput throws an exception, a NullPointerException will be thrown in the finally block. Put these outside the try block

albertshau: also, seems like this can be simplified to:
```
try (OutputStreamWriter osw = new OutputStreamWriter(urlConn.getOutputStream(), "UTF-8")) {
  osw.write(body);
}
```

close() will flush the stream before closing it.

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -0,0 +1,145 @@
...
+        return executeGetCall(getHttpUrlConnection(connectionUrl));
+      } else if (method.equals(HttpMethod.POST) || method.equals(HttpMethod.PUT)) {
+        return executeCallWithBody(getHttpUrlConnection(connectionUrl), method, requestBody);
+      }
+      throw new IllegalStateException("The given Http Method Doesn't match with available methods");
+    });
+
+  }
+
+  private String executeGetCall(HttpURLConnection urlConn) throws IOException {
```
albertshau: as commented elsewhere, this should not be reading the response as a string and then converting the string into an object. Instead, it should convert directly to the object with the Reader.

There also needs to be a separate method thats used when the response is bytes and not json.

delta-app/src/main/java/io/cdap/delta/store/DBReplicationOffsetStore.java
albertshau: did the implementation change? Now it looks like it can't return null.

albertshau: also, this store should be working with the OffsetAndSequence object, not with strings.

Objects should only be encoded/decoded at the very edges -- in the handler or in the store. That way we ensure that we are dealing with proper representations and not malformed strings.

albertshau: The generation also should be a long instead of a Long, since it shouldn't ever be null. This method should be:

`public OffsetAndSequence getOffset(DeltaWorkerId workerId)`

delta-app/src/main/java/io/cdap/delta/store/DBReplicationOffsetStore.java
```diff
@@ -0,0 +1,105 @@
...
+    throws IOException {
+    List<Field<?>> keys = getKey(namespace, appName, generation, instanceId);
+    Optional<StructuredRow> row = table.read(keys);
+    if (!row.isPresent() || row.get().getString(OFFSET_COL) == null) {
+      return StringUtils.EMPTY;
+    }
+    return row.get().getString(OFFSET_COL);
+  }
+
+  public void writeOffset(String namespace, String appName, Long generation, int instanceId, String data)
```
albertshau: this should be `writeOffset(DeltaWorkerId workerId, OffsetAndSequence offset)`

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
```diff
@@ -0,0 +1,140 @@
...
+    try {
+      StructuredTable table = context.getTable(TABLE_ID);
+      return new DBReplicationStateStore(table);
+    } catch (TableNotFoundException e) {
+      throw new IllegalStateException(String.format(
+        "System table '%s' does not exist. Please check your system environment.", TABLE_ID.getName()), e);
+    }
+  }
+
+  public byte[] getStateData(String namespace, String appName, Long generation, int instanceId, String state)
```
albertshau: similar comment to above, use DeltaWorkerId

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
albertshau: I see, it's in the StateStore interface. The Worker doesn't actually ever use it though does it? Looks like the remote implementation just returns null.

delta-app/src/main/java/io/cdap/delta/store/RemoteStateStore.java
```diff
@@ -0,0 +1,94 @@
...
+  @Override
+  public void writeOffset(DeltaWorkerId id, OffsetAndSequence offset) {
+    String requestBody = GSON.toJson(offset);
+    assessmentServiceClient.retryableApiCall(getMethodPath(id, OFFSET_KEY), HttpMethod.PUT, requestBody);
+  }
+
+  @Override
+  public byte[] readState(DeltaWorkerId id, String key) {
+    String byteString = assessmentServiceClient.retryableApiCall(getMethodPathState(id, key), HttpMethod.GET, null);
+    return byteString.getBytes(StandardCharsets.UTF_8);
```
albertshau: mentioned elsewhere, but this is incorrect.

delta-app/src/main/java/io/cdap/delta/store/RemoteStateStore.java
```diff
@@ -0,0 +1,94 @@
...
+
+  @Override
+  public void writeState(DeltaWorkerId id, String key, byte[] val) {
+    String valString = new String(val, StandardCharsets.UTF_8);
+    assessmentServiceClient.retryableApiCall(getMethodPathState(id, key), HttpMethod.PUT, valString);
+  }
+
+  @Override
+  public Long getLatestGeneration(String namespace, String pipelineName) {
+    return null;
```
albertshau: if this and getWorkerInstances() are never expected to be called, better to throw an UnsupportedOperationException instead of returning null

delta-app/src/test/java/io/cdap/delta/app/PipelineDBStoreTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @AfterClass
+  public static void teardown() throws Exception {
+    removeArtifacts("cdc_db");
+  }
+
+  @Override
+  Long getMaxGenerationNum(ApplicationId appId, String path) throws IOException {
+    Long maxGeneration = Long.parseLong(executeCall(HttpMethod.GET,
+                                                     String.format("v1/context/%s/apps/%s/maxgeneration",
```
albertshau: indentation

delta-app/src/test/java/io/cdap/delta/app/PipelineDBStoreTest.java
```diff
@@ -0,0 +1,122 @@
...
+    String offsetAndSequenceJsonStr = executeCall(HttpMethod.GET, url, null);
+    if (offsetAndSequenceJsonStr.equals("")) {
+      return null;
+    }
+    return GSON.fromJson(offsetAndSequenceJsonStr, OffsetAndSequence.class);
+  }
+
+  @Override
+  PipelineReplicationState getPipelineReplicationState(DeltaWorkerId id, String path) throws IOException {
+    StateRequest stateRequest = new StateRequest(id.getPipelineId().getApp(), "lol");
```
albertshau: lol

delta-app/src/main/java/io/cdap/delta/store/DBReplicationStateStore.java
sahusanket: No this one is used by the `getState` API [ assessment handler ] to track the status of the whole pipeline. 

Sorry, I misread the filename earlier

delta-app/src/main/java/io/cdap/delta/store/StateStoreFactory.java
sahusanket: I have replaced by Enums now.

delta-app/src/test/java/io/cdap/delta/app/PipelineDBStoreTest.java
sahusanket: Sorry, rectified.

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
sahusanket: This is used in Unit test cases [ for the main delta pipeline tests ].
Without these, there is no other way to fetch the max generation from Unit test case, which is necessary to create proper `DeltaWorkerId`

albertshau: This is a public endpoint, it's not good to add something that exposes internals just for a unit test.

In the unit test, can you create the StateStore and look it up from there instead of making a rest call?

sahusanket: Yes, that is being done for the HCFS one, using filesystem . 

But for DB/SQL , we need to fetch it from the actual run's DB. I am not sure how to handle that. 

This is for the test class : `DeltaPipelineStateStoreBaseTest` , where in all the state store dependent methods are abstracted out, and is implemented by HCFS and SQL based separately.

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -38,71 +38,65 @@
...
+import javax.ws.rs.core.MediaType;
 
 /**
  * Class to handle Api calls to System Service
  */
 public class AssessmentServiceClient {
 
-  ServiceDiscoverer serviceDiscoverer;
+  private final ServiceDiscoverer serviceDiscoverer;
+  private RetryPolicy retryPolicy;
```
albertshau: private final

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -38,71 +38,65 @@
...
+  private static final Gson GSON = new Gson();
 
   public AssessmentServiceClient(ServiceDiscoverer context) {
     this.serviceDiscoverer = context;
+    setUpRetryPolicy();
```
albertshau: nit: this looks like it's only called in the constructor, in which case it's a bit more readable just to create it here in the constructor

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -38,71 +38,65 @@
...
   }
 
-  public String retryableApiCall(String connectionUrl, String method, @Nullable String requestBody) {
+  public <T> T retryableApiCall(String connectionUrl, String method, @Nullable String requestBody,
+                                @Nullable Class<T> type) {
```
albertshau: type shouldn't be nullable. Also better for this to be a `Type` rather than `Class`

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -38,71 +38,65 @@
...
 
-  private String executeGetCall(HttpURLConnection urlConn) throws IOException {
+  private <T> T executeGetCall(HttpURLConnection urlConn, Class<T> type) throws IOException {
     urlConn.setRequestMethod(HttpMethod.GET);
-    return validateAndRetrieveContent(urlConn);
+    return validateAndRetrieveContent(urlConn, type);
   }
 
-  private String executeCallWithBody(HttpURLConnection urlConn, String method, String body)
+  private <T> T  executeCallWithBody(HttpURLConnection urlConn, String method, String body, Class<T> type)
```
albertshau: extra space after T

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -38,71 +38,65 @@
...
     OutputStreamWriter osw = null;
+    urlConn.setDoOutput(true);
+    urlConn.setRequestMethod(method);
 
     try {
-      urlConn.setDoOutput(true);
-      urlConn.setRequestMethod(method);
-      os = urlConn.getOutputStream();
-      osw = new OutputStreamWriter(os, "UTF-8");
+      osw = new OutputStreamWriter(urlConn.getOutputStream(), "UTF-8");
```
albertshau: does

```
try (OutputStreamWriter osw = ...) {

}
```

not work? I thought it was an auto-closeable?

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -114,8 +108,20 @@ private String validateAndRetrieveContent(@Nullable HttpURLConnection urlConn) t
                                      getError(urlConn));
     }
 
-    try (Reader reader = new InputStreamReader(urlConn.getInputStream(), StandardCharsets.UTF_8)) {
-      return CharStreams.toString(reader);
+    try (InputStream inputStream = urlConn.getInputStream()) {
+      String contentType = urlConn.getContentType();
```
albertshau: it's not good to rely on the content type to determine the return type. Should always use whatever type is passed in.

For bytes, there should be a separate method with a fixed byte[] return type. You can still share most of the code by doing something like:

```

private <T> T retrieve(HttpURLConnection urlConn, Function<HttpURLConnection, T> decodeFunc) {

}

private <T> T retrieve(HttpURLConnection urlConn, Type type) {
  return retrieve(urlConn, conn -> { // use gson to decode })
}

private byte[] retrieve(HttpURLConnection urlConn) {
  return retrieve(urlConn, conn -> { // use BytesStreams to decode })
}
```

delta-app/src/main/java/io/cdap/delta/store/DBStateStoreService.java
```diff
@@ -39,41 +37,28 @@ public DBStateStoreService(TransactionRunner transactionRunner) {
   @Override
   public OffsetAndSequence readOffset(DeltaWorkerId id) {
     return TransactionRunners.run(txRunner, context -> {
-      String offsetStr =  DBReplicationOffsetStore.get(context).getOffsets(id.getPipelineId().getNamespace(),
-                                                              id.getPipelineId().getApp(),
-                                                              id.getPipelineId().getGeneration(),
-                                                              id.getInstanceId());
-      return offsetStr.equals("") ? null : GSON.fromJson(offsetStr, OffsetAndSequence.class);
+      return   DBReplicationOffsetStore.get(context).getOffsets(id);
```
albertshau: remove extra spaces.

This isn't trigger checkstyle failures?

delta-app/src/main/java/io/cdap/delta/store/RemoteStateStore.java
```diff
@@ -49,45 +49,44 @@ public RemoteStateStore(AssessmentServiceClient assessmentServiceClient) {
 
   @Override
   public OffsetAndSequence readOffset(DeltaWorkerId id) {
-    String offsetAndSequenceJsonStr =
-      assessmentServiceClient.retryableApiCall(getMethodPath(id, OFFSET_KEY), HttpMethod.GET, null);
-    return GSON.fromJson(offsetAndSequenceJsonStr, OffsetAndSequence.class);
+    return  assessmentServiceClient
```
albertshau: remove extra spaces

delta-app/src/main/java/io/cdap/delta/store/RemoteStateStore.java
```diff
@@ -49,45 +49,44 @@ public RemoteStateStore(AssessmentServiceClient assessmentServiceClient) {
...
   public void writeState(DeltaWorkerId id, String key, byte[] val) {
     String valString = new String(val, StandardCharsets.UTF_8);
-    assessmentServiceClient.retryableApiCall(getMethodPathState(id, key), HttpMethod.PUT, valString);
+    assessmentServiceClient.retryableApiCall(getMethodPathState(id, key), HttpMethod.PUT, valString, null);
   }
 
   @Override
   public Long getLatestGeneration(String namespace, String pipelineName) {
-    return null;
+    throw new UnsupportedOperationException(" getLatestGeneration is not supposed to be called from a worker instance");
```
albertshau: remove extra space at the start

delta-app/src/main/java/io/cdap/delta/store/RemoteStateStore.java
```diff
@@ -49,45 +49,44 @@ public RemoteStateStore(AssessmentServiceClient assessmentServiceClient) {
...
   }
 
   @Override
   public Collection<Integer> getWorkerInstances(DeltaPipelineId pipelineId) {
-    return null;
+    throw new UnsupportedOperationException(" getWorkerInstances is not supposed to be called from a worker instance");
```
albertshau: remove extra space at the start

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
albertshau: Are you unable to create it the same way its created in the StoreTest for some reason?

delta-app/src/main/java/io/cdap/delta/store/DBStateStoreService.java
sahusanket: No, build was successful .

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
sahusanket: Yes it works. I missed it the first time. 
Rectified it

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
sahusanket: Changed the implementation based on the given type into different functions for bytes/object , reusing most of the code 

I was not able to understand the advantage of using the functional interface here as suggested, We can further discuss if the current one doesn't look right.

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -96,37 +92,41 @@ public AssessmentServiceClient(ServiceDiscoverer context) {
...
    */
-  private <T> T validateAndRetrieveContent(@Nullable HttpURLConnection urlConn, Class<T> type) throws IOException {
+  private <T> T validateAndRetrieveContent(@Nullable HttpURLConnection urlConn, Type type) throws IOException {
     String serviceName = "State Store (AssessorService)";
     if (urlConn == null) {
       throw new RetryableException(serviceName + " service is not available");
     }
 
+    if (type == null) {
+      throw new NullPointerException(" Type parameter cannot be null");
```
albertshau: remove space before Type

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -96,37 +92,41 @@ public AssessmentServiceClient(ServiceDiscoverer context) {
...
-        }
-
-        case MediaType.APPLICATION_OCTET_STREAM: {
-          return (T) ByteStreams.toByteArray(inputStream);
-        }
-
-        default:
-          return (T) CharStreams.toString(new InputStreamReader(inputStream, StandardCharsets.UTF_8));
-      }
+      return type == byte[].class ?  (T) retrieve(inputStream) :  retrieve(inputStream, type);
```
albertshau: should not be type checking here, there should be a method that returns a byte[] and doesn't take any Type. The one that takes a Type and returns T should always be for a json response (technically you can represent a byte[] in json as a literal json array of numbers).

Think of the byte[] as the method used to get the raw byte content of the response, while the Type one is used to get and decode json response.

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -96,37 +92,41 @@ public AssessmentServiceClient(ServiceDiscoverer context) {
...
       urlConn.disconnect();
     }
   }
 
+  private byte[] retrieve(InputStream inputStream) throws IOException {
+    return ByteStreams.toByteArray(inputStream);
+  }
+
+  private <T> T retrieve(InputStream inputStream, Type type) throws IOException {
+    if (type == String.class) {
```
albertshau: don't need this, gson should work on String.class

delta-app/src/main/java/io/cdap/delta/store/RemoteStateStore.java
```diff
@@ -68,17 +67,17 @@ public void writeOffset(DeltaWorkerId id, OffsetAndSequence offset) {
   @Override
   public void writeState(DeltaWorkerId id, String key, byte[] val) {
     String valString = new String(val, StandardCharsets.UTF_8);
-    assessmentServiceClient.retryableApiCall(getMethodPathState(id, key), HttpMethod.PUT, valString, null);
+    assessmentServiceClient.retryableApiCall(getMethodPathState(id, key), HttpMethod.PUT, valString, String.class);
```
albertshau: As commented before, do not try to represent a byte array as a String. It is incorrect. For example, there are some byte arrays that are invalid UTF8 Strings.

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
sahusanket: Actually I tried it out , it was failing. 

`GSON.fromJson(reader, type)`  

**case1** : reader's content is proper **JSON** , type is an **Object** --> Works 

**case2** : reader's content is proper **JSON** , type is an **String.class** --> **Fails**

	Error    `Expected a string but was BEGIN_OBJECT at line 1 column 2` 

**case3** : reader's content is a **string** , type is an **String.class** --> **Fails**

	Error : `com.google.gson.stream.MalformedJsonException`

	A workaround for case3 was using "JsonReader.setLenient(true)" , but it has it's own issues like it returns 1 word 
         , and have to iterate to get whole sentence.

albertshau: not sure what content you're trying to deserialize, but:

```
    String x = "xyz";
    String y = new Gson().fromJson(new StringReader(x), String.class);
```
this works fine for me. It's a bit of a moot point, you shouldn't be reading anything as a string anyway. Only thing being deserialized is OffsetAndSequence.

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -56,80 +57,76 @@ public AssessmentServiceClient(ServiceDiscoverer context) {
...
-   * Copied from "AbstractServiceRetryableMacroEvaluator"
-   * Returns the full content of the error stream for the given {@link HttpURLConnection}.
-   */
-  private <T> T validateAndRetrieveContent(@Nullable HttpURLConnection urlConn, Type type) throws IOException {
-    String serviceName = "State Store (AssessorService)";
-    if (urlConn == null) {
-      throw new RetryableException(serviceName + " service is not available");
+  private <T> T executeCall(HttpURLConnection urlConn, String method, String body, Type type) throws IOException {
+    urlConn.setRequestMethod(method);
+    if (body != null) {
```
albertshau, line 109: if body can be null, the argument should be annotated with `@Nullable`

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -56,80 +57,76 @@ public AssessmentServiceClient(ServiceDiscoverer context) {
...
+      urlConn.setDoOutput(true);
+      try (OutputStreamWriter osw = new OutputStreamWriter(urlConn.getOutputStream(), "UTF-8");) {
```
albertshau: nit: use StandardCharsets.UTF_8 instead of hardcoded string

delta-app/src/main/java/io/cdap/delta/store/RemoteStateStore.java
```diff
@@ -0,0 +1,91 @@
...
+  @Override
+  public OffsetAndSequence readOffset(DeltaWorkerId id) {
+    return assessmentServiceClient
+      .retryableApiCall(getMethodPath(id, OFFSET_KEY), HttpMethod.GET, null, OffsetAndSequence.class);
+  }
+
+  @Override
+  public void writeOffset(DeltaWorkerId id, OffsetAndSequence offset) {
+    String requestBody = GSON.toJson(offset);
+    assessmentServiceClient.retryableApiCall(getMethodPath(id, OFFSET_KEY), HttpMethod.PUT, requestBody, String.class);
```
albertshau: Type here shouldn't be passed in since you don't care about the response body. There should be one with void return type that doesn't take a Type argument.

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
sahusanket: Yes this works, but StringReader takes only `String` in it's constructor . We wanted to avoid conversion from bytes[] to string for normal object creation. 

But yes, as of now we only convert `OffsetAndSequence` ,  I will remove the String part for now, wanted to keep it generic before.

delta-app/src/main/java/io/cdap/delta/store/AssessmentServiceClient.java
```diff
@@ -91,11 +90,17 @@ public AssessmentServiceClient(ServiceDiscoverer context) {
     return failsafe.get(() ->  executeCall(getHttpUrlConnection(connectionUrl), method, requestBody, type));
   }
 
-  private <T> T executeCall(HttpURLConnection urlConn, String method, String body, Type type) throws IOException {
+  public void retryableApiCall(String connectionUrl, String method, @Nullable String requestBody) {
+    //Calls to Post/Put a body in Assessment Service will return only Status Code [ INTEGER ].
+    retryableApiCall(connectionUrl, method, requestBody, Integer.class);
```
albertshau: The methods return the status code as the response body? That seems odd.

It's better not to read the response body at all if it won't be used. For example, pass null as the Type and skip the decoding if Type is null.

delta-app/src/main/java/io/cdap/delta/store/RemoteStateStore.java
```diff
@@ -55,13 +54,13 @@ public OffsetAndSequence readOffset(DeltaWorkerId id) {
...
     String requestBody = GSON.toJson(offset);
-    assessmentServiceClient.retryableApiCall(getMethodPath(id, OFFSET_KEY), HttpMethod.PUT, requestBody, String.class);
+    assessmentServiceClient.retryableApiCall(getMethodPath(id, OFFSET_KEY), HttpMethod.PUT, requestBody);
   }
 
   @Override
   public byte[] readState(DeltaWorkerId id, String key) {
     return assessmentServiceClient
-      .retryableApiCall(getMethodPathState(id, key), HttpMethod.GET, null);
+      .retryableApiCall(getMethodPathState(id, key), HttpMethod.GET, new byte[0]);
```
albertshau: it's weird to pass a body when doing a GET and there is no request body. Think it's a bit more clear what's happening to keep the body as null.

albertshau (approved):
minor comments otherwise lgtm. please fix and squash before merging.

sagarkapare (approved):
last commit LGTM.

sahusanket:
Build link : https://builds.cask.co/browse/DEL-BT159-2

PR #155, opened 2021-10-28
add severity for a problem

seanzhougoogle:


delta-api/src/main/java/io/cdap/delta/api/assessment/Problem.java
```diff
@@ -26,12 +26,18 @@
   private final String description;
   private final String suggestion;
   private final String impact;
+  private final Severity severity;
```
sagarkapare, line 29: We should include this field in the `equals()` and `hashCode()`

sagarkapare (approved):
One comment. else lgtm.

delta-api/src/main/java/io/cdap/delta/api/assessment/Problem.java
seanzhougoogle: good catch, done :)

PR #159, opened 2021-11-10
Add transformation validation and support multiple columns using same directive

seanzhougoogle:
1. add transformation validations
2. support multiple column using same directive

### DIFF
--- a/delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
+++ b/delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
@@ -197,7 +197,6 @@ public class DeltaConfig extends Config {
       for (TableTransformation tableTransformation : tableTransformations) {
         tableTransformation.validate();
       }
-
     }
     return Optional.ofNullable(targetStage);
   }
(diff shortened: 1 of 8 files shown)


delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
```diff
@@ -191,6 +192,13 @@ public Stage getTarget() {
     if (sourceStage == null) {
       throw new IllegalArgumentException("No source found.");
     }
+
+    if (tableTransformations != null) {
+      for (TableTransformation tableTransformation : tableTransformations) {
+        tableTransformation.validate();
+      }
+
```
sagarkapare: minor nit: remove extra line.

delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
```diff
@@ -191,6 +192,13 @@ public Stage getTarget() {
     if (sourceStage == null) {
       throw new IllegalArgumentException("No source found.");
     }
+
+    if (tableTransformations != null) {
```
sagarkapare, line 196: I think I missed in last PR review, but we should add `tableTransformations` to the equals and hashCode method as well.

sagarkapare (commented):
minor comments, else LGTM 👍

sagarkapare (approved):

delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
seanzhougoogle: done :)

delta-proto/src/main/java/io/cdap/delta/proto/DeltaConfig.java
seanzhougoogle: good catch :) Added equals and hashCode method for TableTransformations and ColumnTransformations too :)

PR #163, opened 2021-12-16
CDAP-18670 : migration utility from hcfs to sql based state store

sahusanket:


delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -159,10 +160,14 @@ protected void configure() {
...
     super.initialize(context);
     long generation = Long.parseLong(context.getSpecification().getProperty(GENERATION));
-
     ApplicationSpecification appSpec = context.getApplicationSpecification();
+    StateStore stateStore = new RemoteStateStore(context);
+    DeltaWorkerId id = new DeltaWorkerId(new DeltaPipelineId(context.getNamespace(), appSpec.getName(), generation),
+                                         context.getInstanceId());
     config = GSON.fromJson(appSpec.getConfiguration(), DeltaConfig.class);
 
+    new StateStoreMigrator(config.getOffsetBasePath(), id, stateStore).perform();
```
sagarkapare: minor nit: Assign to local variable and then do var.perform();

delta-app/src/main/java/io/cdap/delta/store/StateStoreMigrator.java
```diff
@@ -0,0 +1,116 @@
...
+
+    //Initiate HCFS state store to read
+    Path path = new Path(hcfsStateStorePath);
+    hcfsStateStore = HCFSStateStore.from(path);
+
+    //Get previous latest generation : required as we cannot access previous context.
+    Long previousGeneration = hcfsStateStore.getLatestGeneration(currentId.getPipelineId().getNamespace(),
+                                                                 currentId.getPipelineId().getApp());
+    if (previousGeneration == null) {
+      return;
```
sagarkapare: When can previousGeneration be null? Maybe worth logging as DEBUG log.

delta-app/src/main/java/io/cdap/delta/store/StateStoreMigrator.java
```diff
@@ -0,0 +1,116 @@
...
+
+    fileSystem.rename(instanceBasePath, new Path(path, String.format("%s/%s/%d/%s",
+                                                                     previousId.getPipelineId().getNamespace(),
+                                                                     previousId.getPipelineId().getApp(),
+                                                                     previousId.getPipelineId().getGeneration(),
+                                                                     previousId.getInstanceId() + "_migrated")));
+  }
+
+  private void migrateOffset() throws IOException {
+    if (remoteStateStore.readOffset(this.currentId) == null) {
```
sagarkapare, line 110: Do we need this if condition? migration will only happen once right? after that files are renamed so migrateOffset should not get called again?

delta-app/src/main/java/io/cdap/delta/store/StateStoreMigrator.java
```diff
@@ -0,0 +1,116 @@
...
+      remoteStateStore.writeOffset(this.currentId, hcfsStateStore.readOffset(this.previousId));
+    }
+  }
+
+  private void migrateStateFile(String key) throws IOException {
+    byte[] presentDBStateData = remoteStateStore.readState(this.currentId, key);
+    if (presentDBStateData == null || presentDBStateData.length == 0) {
```
sagarkapare, line 118: similarly here.. why do we need to check present state of the store?

delta-app/src/main/java/io/cdap/delta/store/StateStoreMigrator.java
```diff
@@ -0,0 +1,116 @@
...
+      return;
+    }
+
+    for (FileStatus stateFile : fStats) {
+      Path filePath = stateFile.getPath();
+
+      if (filePath.getName().equals(RemoteStateStore.OFFSET_KEY)) {
+        migrateOffset();
+      } else {
+        migrateStateFile(filePath.getName());
```
sagarkapare, line 96: Also better to log what are are migrating as DEBUG.

delta-app/src/main/java/io/cdap/delta/store/StateStoreMigrator.java
```diff
@@ -0,0 +1,116 @@
...
+  protected HCFSStateStore hcfsStateStore;
+  protected DeltaWorkerId previousId;
+
+  public StateStoreMigrator(String hcfsStateStorePath, DeltaWorkerId currentId, StateStore remoteStateStore) {
+    this.hcfsStateStorePath = hcfsStateStorePath;
+    this.currentId = currentId;
+    this.remoteStateStore = remoteStateStore;
+  }
+
+  public void perform() throws IOException {
```
sagarkapare: Maybe better to return boolean from this method to indicate whether we did any migration as a part of this call. in worker where we call this method then we can log the output as DEBUG.

sagarkapare (commented):
minor nits around adding more DEBUG logging, else looks good 👍

delta-app/src/main/java/io/cdap/delta/store/StateStoreMigrator.java
sahusanket: It is only possible when user manually intervenes with the GCS path. It's rare case, but if it happens There will NPE in the following code. 
And anyway if it's null or not present, there won;t be migration and it's not worth throwing NPE rather we can skip and continue . 

Yes I will add debug logs

delta-app/src/main/java/io/cdap/delta/store/StateStoreMigrator.java
sahusanket: So this is to mitigate the case when 
- We have upserted data into SQL , but before the folder is renamed there is a failure. (although in next try , the same data should be re written) 
- Also in future incase the folder folder is renamed back to original name, then on restart it might overwrite the newer values in DB. 

It's just a safety check to enforce only write once.

sagarkapare (approved):

PR #164, opened 2022-01-20
Cherry pick transformation framework

seanzhougoogle:


delta-api/pom.xml
```diff
@@ -36,6 +36,12 @@
       <artifactId>cdap-api</artifactId>
       <version>${cdap.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-api-common</artifactId>
+      <version>6.6.0-SNAPSHOT</version>
```
sagarkapare: `${cdap.version}`?

delta-api/pom.xml
```diff
@@ -36,6 +36,12 @@
...
+      <scope>compile</scope>
```
sagarkapare: I think `compile` is default scope, you dont need to specify explicitly.

sagarkapare (approved):
LGTM on cherry-pick assuming no conflicts.

delta-api/pom.xml
seanzhougoogle: done :)

delta-api/pom.xml
seanzhougoogle: done :)

PR #165, opened 2022-01-21
fix unit test

yaojiefeng:
This is dependent on https://github.com/cdapio/cdap/pull/13850, added the logic to put initial generation time for the runtime deploy so the generation time won't get overriden. Verified PipelineDBStoreTest is passing now.


fernst (approved):
LGTM

PR #166, opened 2022-02-07
CDAP-18811 update the error message

seanzhougoogle:


delta-proto/src/main/java/io/cdap/delta/proto/ColumnTransformation.java
```diff
@@ -43,10 +43,12 @@ public String getDirective() {
    */
   public void validate() {
     if (columnName == null || columnName.isEmpty()) {
-      throw new IllegalArgumentException("Column name of a column transformation should not be null or empty.");
+      throw new IllegalArgumentException(String.format("Column name of a ColumnTransformation for directive %s is " +
+                                                         "null or empty.", directive));
```
sagarkapare, line 47: seems like `directive` name can also be null or empty in which case this message will be confusing. Since we have this hierarchy table -> column -> directive in the configuration, it would be good to log table name for column validation error and table name and column name for directive validation errors.

sagarkapare: We can catch the IllegalArgumentException in the caller of validate and add table name as an additional context to exception message. Something like in `TableTransformation` class

```java
  public void validate() {
    if (tableName == null || tableName.isEmpty()) {
      throw new IllegalArgumentException(String.format("Table name of a TableTransformation that contains below " +
                                                         "ColumnTransformations : %s is null or empty.",
                                                       columnTransformations));
    }
    if (columnTransformations != null) {
      for (ColumnTransformation columnTransformation : columnTransformations) {
        try {
          columnTransformation.validate();
        } catch (IllegalArgumentException e) {
          throw new IllegalArgumentException(String.format("Column/Directive validation failed for table %s." +
                                                             " Reason: %s", tableName, e.getMessage()));
        }
      }
    }
} 
```

delta-proto/src/main/java/io/cdap/delta/proto/ColumnTransformation.java
```diff
@@ -67,5 +69,10 @@ public boolean equals(Object o) {
   public int hashCode() {
     return Objects.hash(columnName, directive);
   }
+
+  @Override
+  public String toString() {
+    return String.format("ColumnTransformation(columnName : \"%s\", directive : \"%s\")", columnName, directive);
```
sagarkapare, line 75: should we have `null` checks for columnName and directive, since both are `nullable`

delta-proto/src/main/java/io/cdap/delta/proto/TableTransformation.java
```diff
@@ -49,7 +49,9 @@ public String getTableName() {
    */
   public void validate() {
     if (tableName == null || tableName.isEmpty()) {
-      throw new IllegalArgumentException("Table name of a TableTransformation should not be null or empty");
+      throw new IllegalArgumentException(String.format("Table name of a TableTransformation that contains below " +
+                                                         "ColumnTransformations : %s is null or empty.",
+                                                       columnTransformations));
```
sagarkapare: `columnTransformations` can be `null` so would be good to check that here.

sagarkapare (commented):
few minor comments.

delta-proto/src/main/java/io/cdap/delta/proto/ColumnTransformation.java
seanzhougoogle: that's a good idea:)

seanzhougoogle: done :)

delta-proto/src/main/java/io/cdap/delta/proto/ColumnTransformation.java
seanzhougoogle: for `toString` method, it should be fine, if it's null , then it prints `null` in the string.

delta-proto/src/main/java/io/cdap/delta/proto/TableTransformation.java
seanzhougoogle: hmm... don't quite understand, if it's null, what should we print in the error message ?

seanzhougoogle: discussed offline, will remove the columnTransformations in the message here.

sagarkapare (approved):

PR #169, opened 2022-02-13
CDAP-18852 : adding on the fly draft assessment

sahusanket:


delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -166,6 +166,27 @@ public void assessDraft(HttpServiceRequest request, HttpServiceResponder respond
...
   }
 
+  @POST
+  @Path("v1/contexts/{context}/assessPipeline")
+  public void assessInterimDraft(HttpServiceRequest request, HttpServiceResponder responder,
+                          @PathParam("context") String namespaceName) {
+    respond(namespaceName, responder, ((namespace) -> {
+      Draft draft;
+      try {
+        draft = GSON.fromJson(StandardCharsets.UTF_8.decode(request.getContent()).toString(), Draft.class);
```
sagarkapare: Should we simply accept `DeltaConfig` in this API? Rest of the fields such as draft name, label etc. can be generated by backend. 

This way there is a cleaner separation between the two APIs

1. When you have draft, use
`v1/contexts/{context}/drafts/{draft}/assessPipeline`

2. When you don't have draft but the config, use
`v1/contexts/{context}/assessPipeline`

sagarkapare (commented):
One comment about API design

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
sahusanket: Thanks, sounds good.

Made the changes and did some refactoring to have methods free from draft terms, so we can reuse them without creating a draft / draftid

sagarkapare (approved):
LGTM. Thanks for fix!

PR #170, opened 2022-02-18
CDAP-18867 : Adding detailed error response as a part of assessment a…

sahusanket:
…pi calls for replication transformation errors

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/store/DraftService.java
+++ b/delta-app/src/main/java/io/cdap/delta/store/DraftService.java
@@ -50,12 +50,16 @@ import io.cdap.delta.proto.TableTransformation;
 import io.cdap.transformation.ColumnRenameInfo;
 import io.cdap.transformation.DefaultMutableRowSchema;
 import io.cdap.transformation.DefaultRenameInfo;
+import io.cdap.transformation.TransformationException;
 import io.cdap.transformation.TransformationUtil;
 import io.cdap.transformation.api.Transformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -69,6 +73,7 @@ import javax.annotation.Nullable;
  * Handles logic around storage, retrieval, and assessment of drafts.
  */
 public class DraftService {
+  private static final Logger LOG = LoggerFactory.getLogger(DraftService.class);
   private final TransactionRunner txRunner;
   private final PropertyEvaluator propertyEvaluator;
 
@@ -229,16 +234,30 @@ public class DraftService {
 
     Map<String, TableTransformation> tableLevelTransformations =
       TransformationUtil.getTableLevelTransformations(deltaConfig);
-    List<Transformation> transformations = TransformationUtil.loadTransformations(configurer,
-                                                                                  tableLevelTransformations,
-                                                                                  selectedTable);
+    Map<String, List<Transformation>> transformations = new HashMap<>();
+    try {
+      transformations = TransformationUtil.loadTransformations(configurer, tableLevelTransformations, selectedTable);
+    } catch (TransformationException e) {
+      LOG.debug("Failed to apply transformation: ", e);
+      List<Problem> transformationErrors = new ArrayList<>();
+      transformationErrors.add(new Problem("Transformation Loading failed",
+                               String.format("Failed to load transformations for the table : %s and column : %s " +
+                               "with error : %s", e.getTableName(), e.getColumnName(), e.getMessage()),
+                               "Please ensure the applied transformation's plugin is uploaded'",
+                               ""));
+
+      return new TableAssessmentResponse(Collections.emptyList(), Collections.emptyList(), transformationErrors);
+    }
 
     try (TableRegistry tableRegistry = createTableRegistry(draftId.getNamespace(), deltaConfig, configurer);
          TableAssessor<TableDetail> sourceTableAssessor = createTableAssessor(configurer, deltaConfig.getSource(),
            deltaConfig.getTables());
          TableAssessor<StandardizedTableDetail> targetTableAssessor = createTableAssessor(configurer, target,
            deltaConfig.getTables())) {
-      return assessTable(selectedTable, tableRegistry, sourceTableAssessor, targetTableAssessor, transformations);
+
+      TableAssessmentResponse assessment = assessTable(selectedTable, tableRegistry, sourceTableAssessor,
+                                                       targetTableAssessor, transformations);
+      return assessment;
     }
   }
 
@@ -281,6 +300,7 @@ public class DraftService {
 
       List<Problem> missingFeatures = new ArrayList<>();
       List<Problem> connectivityIssues = new ArrayList<>();
+      List<Problem> transformationIssues = new ArrayList<>();
       List<TableSummaryAssessment> tableAssessments = new ArrayList<>();
 
       Assessment sourceAssessment = sourceTableAssessor.assess();
@@ -305,13 +325,14 @@ public class DraftService {
         String db = sourceTable.getDatabase();
         String table = sourceTable.getTable();
         try {
-          List<Transformation> transformations = TransformationUtil.loadTransformations(configurer,
+          Map<String, List<Transformation>> transformations = TransformationUtil.loadTransformations(configurer,
                                                                                         tableLevelTransformations,
                                                                                         sourceTable);
           TableAssessmentResponse assessment = assessTable(sourceTable, tableRegistry, sourceTableAssessor,
                                                            targetTableAssessor, transformations);
           missingFeatures.addAll(assessment.getFeatures());
           tableAssessments.add(summarize(db, sourceTable.getSchema(), sourceTable, assessment));
+          transformationIssues.addAll(assessment.getTransformationIssues());
         } catch (TableNotFoundException e) {
           connectivityIssues.add(
             new Problem("Table Not Found",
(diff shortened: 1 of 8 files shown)


delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -229,16 +231,28 @@ public TableAssessmentResponse assessTable(DraftId draftId, Configurer configure
...
+    List<Problem> transformationErrors = new ArrayList<>();
+    try {
+      transformations = TransformationUtil.loadTransformations(configurer, tableLevelTransformations, selectedTable);
+    } catch (TransformationException e) {
+      transformationErrors.add(new Problem("Transformation Loading failed",
+                               String.format("Failed to load transformations for the table : %s and column : %s " +
+                               "with error : %s", e.getTableName(), e.getColumnName(), e.getMessage()),
+                               "Please ensure the applied transformation's plugin is uploaded'",
+                               ""));
+    }
```
seanzhougoogle, line 250: shall we return here ? if any of the transformation fail to be loaded, we don't need to continue

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -229,16 +231,28 @@ public TableAssessmentResponse assessTable(DraftId draftId, Configurer configure
 
     Map<String, TableTransformation> tableLevelTransformations =
       TransformationUtil.getTableLevelTransformations(deltaConfig);
-    List<Transformation> transformations = TransformationUtil.loadTransformations(configurer,
-                                                                                  tableLevelTransformations,
-                                                                                  selectedTable);
+    Map<String, List<Transformation>> transformations = new HashMap<>();
```
seanzhougoogle, line 237: no need to initialize here ?

sahusanket: yes otherwise it gives a compilation error at line 253 .

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
sahusanket: yeah makes sense. Addressed, PTAL

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
sagarkapare: now we can remove the initialization here?

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -229,16 +231,30 @@ public TableAssessmentResponse assessTable(DraftId draftId, Configurer configure
...
+    List<Problem> transformationErrors = new ArrayList<>();
```
sagarkapare: This can be local to catch block below?

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -229,16 +231,30 @@ public TableAssessmentResponse assessTable(DraftId draftId, Configurer configure
...
     try (TableRegistry tableRegistry = createTableRegistry(draftId.getNamespace(), deltaConfig, configurer);
          TableAssessor<TableDetail> sourceTableAssessor = createTableAssessor(configurer, deltaConfig.getSource(),
            deltaConfig.getTables());
          TableAssessor<StandardizedTableDetail> targetTableAssessor = createTableAssessor(configurer, target,
            deltaConfig.getTables())) {
-      return assessTable(selectedTable, tableRegistry, sourceTableAssessor, targetTableAssessor, transformations);
+
+      TableAssessmentResponse assessment = assessTable(selectedTable, tableRegistry, sourceTableAssessor,
+                                                       targetTableAssessor, transformations);
+      assessment.getTransformationIssues().addAll(transformationErrors);
```
sagarkapare: we dont need this now right? since transformationErrors would always be empty here?

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -388,34 +412,36 @@ private TableAssessmentResponse assessTable(SourceTable sourceTable, TableRegist
...
+    return merge(srcAssessment, targetAssessment, columnRenameInfo, transformationIssues);
   }
 
   /**
    * Merge the assessment from the source and target into a single assessment.
    * This amounts to merging the assessment for each column.
    */
   private TableAssessmentResponse merge(TableAssessment srcAssessment, TableAssessment targetAssessment,
-                                        ColumnRenameInfo columnRenameInfo) {
+                                        ColumnRenameInfo columnRenameInfo, List<Problem> transformationProblem) {
```
sagarkapare: better to be consistent with naming `transformationIssues` vs `transformationProblem`

delta-app/src/main/java/io/cdap/transformation/TransformationUtil.java
```diff
@@ -64,6 +67,30 @@ public static String parseDirectiveName(String directive) {
...
   }
 
+  /**
+   * A helper method to reuse transformSchema with column level detail which is to be called during pipeline assessment
+   * to track issues at a column level for better display of errors
+   *
+   * @param schema
+   * @param transformations
+   * @return the transformed mutable row schema.
+   * @throws TransformationException
```
sagarkapare, line 77: please add details about what these parameters are.

delta-app/src/main/java/io/cdap/transformation/TransformationException.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * An exception class to track to which table.column the transformation issue belongs to
+ */
+public class TransformationException extends Exception {
+
+  private final String tableName;
+  private final String columnName;
+
+  public TransformationException(String tableName, String columnName, String message) {
```
sagarkapare: Better to have constructor which takes Throwable as well to preserve it.

delta-app/src/main/java/io/cdap/transformation/TransformationUtil.java
```diff
@@ -64,6 +67,30 @@ public static String parseDirectiveName(String directive) {
...
+  public static DefaultMutableRowSchema transformSchema(String table,
+                                                        Schema schema,
+                                                        Map<String, List<Transformation>> transformations)
+    throws TransformationException {
+    DefaultMutableRowSchema rowSchema = new DefaultMutableRowSchema(schema);
+    for (String columnName : transformations.keySet()) {
+      try {
+        rowSchema = transformSchema(rowSchema, transformations.get(columnName));
+      } catch (Exception e) {
+        throw new TransformationException(table, columnName, e.getMessage());
```
sagarkapare: We are loosing stacktrace from e right?

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -388,34 +412,36 @@ private TableAssessmentResponse assessTable(SourceTable sourceTable, TableRegist
...
        standardizedDetail = new StandardizedTableDetail(standardizedDetail.getDatabase(),
                                                         standardizedDetail.getTable(),
                                                         standardizedDetail.getPrimaryKey(),
                                                         rowSchema.toSchema());
        columnRenameInfo = rowSchema.getRenameInfo();
-      } catch (Exception e) {
-        throw new RuntimeException(String.format("Failed to apply transformations on the schema : %s",
-                                                 e.getMessage()), e);
+      } catch (TransformationException e) {
+        transformationIssues.add(new Problem("Transformation failed",
```
sagarkapare, line 432: Should we log stack trace here maybe as DEBUG?

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -325,16 +343,22 @@ public PipelineAssessment assessPipeline(Namespace namespace, DeltaConfig deltaC
                                       table, db, e.getMessage()),
                         "Check permissions and database connectivity",
                         null));
+        } catch (TransformationException e) {
+          transformationIssues.add(new Problem("Transformation Loading failed",
```
sagarkapare, line 351: Should we log stack trace here as DEBUG?

delta-app/src/main/java/io/cdap/transformation/TransformationUtil.java
```diff
@@ -107,29 +139,35 @@ public static DDLEvent transformDDLEvent(DDLEvent ddlEvent, List<Transformation>
...
+
+        if (columnTransformations.containsKey(columnName)) {
+          columnTransformations.getOrDefault(columnName, new ArrayList<>()).add(transformation);
+        } else {
+          columnTransformations.put(columnName, new ArrayList<>(Arrays.asList(transformation)));
+        }
       } catch (Exception e) {
-        throw new RuntimeException(String.format("Failed to load transformation plugin for directive : %s. Error : %s.",
-                                                 directive, e.getMessage()), e);
+        throw new TransformationException(tableName, columnName, String.format("Exception while loading plugin " +
```
sagarkapare, line 167: Better to preserve `e` here. Also log it as DEBUG from wherever its getting called in the end before returning to the user.

sagarkapare (commented):
Few comments, mainly around logging stack trace as DEBUG or TRACE just before we return and preserving stack trace.

delta-app/src/main/java/io/cdap/transformation/TransformationUtil.java
```diff
@@ -165,7 +165,7 @@ public static DDLEvent transformDDLEvent(DDLEvent ddlEvent, List<Transformation>
         }
       } catch (Exception e) {
         throw new TransformationException(tableName, columnName, String.format("Exception while loading plugin " +
-                                          "for directive : %s. Error : %s.", directive, e.getMessage()));
+                                          "for directive : %s. Error : %s.", directive, e.getMessage()), e.getCause());
```
sagarkapare: just use `e` instead of `e.getCause()` which can be null

delta-app/src/main/java/io/cdap/transformation/TransformationUtil.java
```diff
@@ -85,7 +85,7 @@ public static DefaultMutableRowSchema transformSchema(String table,
       try {
         rowSchema = transformSchema(rowSchema, transformations.get(columnName));
       } catch (Exception e) {
-        throw new TransformationException(table, columnName, e.getMessage());
+        throw new TransformationException(table, columnName, e.getMessage(), e.getCause());
```
sagarkapare: same here. pass `e` instead of `e.getCause()`

delta-app/src/main/java/io/cdap/delta/store/DraftService.java
```diff
@@ -232,10 +235,11 @@ public TableAssessmentResponse assessTable(DraftId draftId, Configurer configure
     Map<String, TableTransformation> tableLevelTransformations =
       TransformationUtil.getTableLevelTransformations(deltaConfig);
     Map<String, List<Transformation>> transformations = new HashMap<>();
-    List<Problem> transformationErrors = new ArrayList<>();
     try {
       transformations = TransformationUtil.loadTransformations(configurer, tableLevelTransformations, selectedTable);
     } catch (TransformationException e) {
+      LOG.debug("Transformation failure : " + e);
```
sagarkapare: I think `"Transformation failure : " + e` will call e.toString() which could just be reference to the exception. Use `LOG.debug("Failed to apply transformation: ", e)` format everywhere in this commit.

sagarkapare (approved):
few final comments. else LGTM 👍

PR #188, opened 2022-09-13
CDAP-19589 - Add sort keys to DML event for reliable ordering of events

sumitjnn:
- Updated unit tests
- Verified no checkstyle errors


sahusanket (changes requested):
We will not add sort keys to DDL events?

sumitjnn:
> We will not add sort keys to DDL events?

No, sort keys are not required for DDL events as they are applied immediately and do not require ordering.

seanzhougoogle (approved):

seanzhougoogle:
> > We will not add sort keys to DDL events?
> 
> No, sort keys are not required for DDL events as they are applied immediately and do not require ordering.

Mainly it's because DDL events are not generated from Datastream, we generate the DDL events in order ourselves.

sahusanket (approved):

PR #199, opened 2022-12-06
CDAP-20113 - Offload plugin execution to Task Workers in security enabled configurations

sumitjnn:
https://cdap.atlassian.net/browse/CDAP-20113 

Assessment logic which invokes plugins is extracted out of `DraftService` and moved to a separate class `Assessor` which can also be used by remote tasks. Another reason is that we don't want to use `DraftService` in remote task as it makes DB calls. Extracted macro evaluation logic to separate class so that it can be invoked before sending remote task.

### Done
- Unit tests
- Sanity/happy scenarios on security enabled instance
- Error scenarios on security enabled instance. Verified stack trace from task worker is logged.
- Sanity/happy scenarios on security enabled instance
- Error scenarios on security enabled instance


delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
```diff
@@ -161,10 +192,18 @@ public void assessDraft(HttpServiceRequest request, HttpServiceResponder respond
                           @PathParam("draft") String draftName) {
     respond(namespaceName, responder, ((namespace) -> {
       DraftId draftId = new DraftId(namespace, draftName);
-      PluginConfigurer pluginConfigurer = getContext().createPluginConfigurer(namespaceName);
-      PipelineAssessment assessment =
-        getDraftService().assessPipeline(draftId, new DefaultConfigurer(pluginConfigurer));
-      responder.sendString(GSON.toJson(assessment));
+      Draft draft = getDraftService().getDraft(draftId);
+
+      draft.getConfig().validatePipeline();
```
sumitjnn, line 197: Moved here since its better to fail validation early before sending task to worker

delta-app/src/main/java/io/cdap/delta/app/service/Assessor.java
```diff
@@ -0,0 +1,475 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
+/**
+ * Handles logic around assessment of delta config.
+ */
+public class Assessor {
```
sumitjnn, line 74: Assessment logic which invokes plugins is extracted out of DraftService and moved to a separate class which can also be used by remote tasks. We don't want to use DraftService in remote task as it also makes DB calls.

The code in this class is mostly refactored as is from DraftService with slight change in parameters, refer https://gist.github.com/sumitjnn/47cc2a65a4e6730d3337f8a0180264b6/revisions for comparison with the old code in draft service

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
dli357: Does the draft validation perform any validation which may run plugin-specific code? I traced some of the code and noticed this calls into the `Plugin#validate()` method, but I'm unclear on whether that could be overridden or not given it's in the proto module.

delta-app/src/main/java/io/cdap/delta/app/service/Assessor.java
dli357: By DB calls, do you mean calls to the user-provided database, or calls to the CDAP's structured table DB?

delta-app/src/main/java/io/cdap/delta/macros/MacroEvaluator.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public class MacroEvaluator {
+ private final PropertyEvaluator propertyEvaluator;
+
+ public MacroEvaluator(PropertyEvaluator propertyEvaluator) {
+  this.propertyEvaluator = propertyEvaluator;
+ }
+
+ public DeltaConfig evaluateMacros(Namespace namespace, DeltaConfig config) {
+  DeltaConfig.Builder builder = DeltaConfig.builder()
```
dli357: Can we create a helper method in the DeltaConfig builder which takes an existing DeltaConfig so that the copying takes place in a single location? If we add fields in the future to the config, this could be easy to lose track of.

delta-app/src/main/java/io/cdap/delta/macros/PropertyEvaluator.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2020 Cask Data, Inc.
+ * Copyright © 2022 Cask Data, Inc.
```
dli357: nit: copyrights should be start year to current year, i.e. `2020-2022`

delta-app/src/main/java/io/cdap/delta/macros/MacroEvaluator.java
```diff
@@ -0,0 +1,59 @@
...
+import io.cdap.delta.proto.Plugin;
+import io.cdap.delta.proto.Stage;
+import io.cdap.delta.store.Namespace;
+
+import java.util.Map;
+
+/**
+ * Evaluates macros in delta config
+ */
+public class MacroEvaluator {
```
dli357: nit: Maybe we should call this ConfigMacroEvaluator since it's intended to evaluate macros specifically in the config?

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
sumitjnn: I think it should not run any plugin specific code, as `Plugin` is just a metadata class to represent it's properties in the draft or app spec https://github.com/data-integrations/delta/blob/develop/delta-proto/src/main/java/io/cdap/delta/proto/Plugin.java

delta-app/src/main/java/io/cdap/delta/app/service/Assessor.java
sumitjnn: Its CDAP's structured table DB which stores drafts

delta-app/src/main/java/io/cdap/delta/macros/MacroEvaluator.java
sumitjnn: Makes sense, changed

delta-app/src/main/java/io/cdap/delta/macros/PropertyEvaluator.java
sumitjnn: Thanks, fixed

delta-app/src/main/java/io/cdap/delta/app/service/AssessmentHandler.java
dli357: Sounds good, thanks!

dli357 (approved):
Thanks!

PR #203, opened 2022-12-28
CDAP-19622 : using generation number from app spec from app configurator

sahusanket:


delta-app/src/main/java/io/cdap/delta/app/DeltaWorker.java
```diff
@@ -146,8 +145,8 @@ protected void configure() {
 
     // if this is runtime deploy, add the generation time at initial deploy since the offset is dependent on
```
sumitjnn: Please update the comment as well

sumitjnn (commented):
Please add a unit test

sumitjnn (approved):

PR #219, opened 2023-03-16
CDAP-19929 - Add toString() method for DDL Event and DDL operation for logging

yjhawar:


delta-api/src/main/java/io/cdap/delta/api/DDLEvent.java
```diff
@@ -69,6 +71,22 @@ public boolean equals(Object o) {
...
+    String primaryKeyString = primaryKey == null || primaryKey.isEmpty() ?
+                              null : "primaryKey=" + "[" + String.join(",", primaryKey) + "]";
+    String offsetString = this.getOffset().get() != null ? "offset=" + this.getOffset().get() : null;
+    String sourceTimeString  = this.getSourceTimestampMillis() != null ? "sourceTimestampMillis=" +
+                              this.getSourceTimestampMillis() : null;
+    String ddlEvent = Stream.of(changeTypeString, operationString, primaryKeyString, offsetString,
+                                    isSnapshotString, sourceTimeString)
+                            .filter(s -> s != null && !s.isEmpty())
+                            .collect(Collectors.joining(", "));
+    return "{" + ddlEvent + "}";
```
sumitjnn: 1. We can use StringJoiner instead which supports delimeter, oprefix and suffix
2. We should implement toString in Offset class instead of fetching the map here
3. We can directly rely on List's toString for primary key

```suggestion
    StringJoiner stringJoiner = new StringJoiner(",", "{", "}");
    stringJoiner.add("operation=" + operation);
    stringJoiner.add("isSnapshot=" + this.isSnapshot());
    stringJoiner.add("changeType=" + this.getChangeType());
    if (primaryKey != null && !primaryKey.isEmpty()) {
      stringJoiner.add("primaryKey=" + primaryKey);
    }
    if (this.getOffset() != null) {
      stringJoiner.add("offset=" + this.getOffset());
    }
    if (this.getSourceTimestampMillis() != null) {
      stringJoiner.add("sourceTimestampMillis=" + this.getSourceTimestampMillis());
    }
    return stringJoiner.toString();
```

delta-api/src/main/java/io/cdap/delta/api/DDLOperation.java
```diff
@@ -114,4 +116,17 @@ public boolean equals(Object o) {
...
+  public String toString() {
+    String database = "databaseName=" + databaseName;
+    String operationType = "type=" + type;
+    String schema = schemaName != null ? "schemaName=" + schemaName : null;
+    String table = tableName != null ? "tableName=" + tableName : null;
+    String prevTable = prevTableName != null ? "prevTableName=" + prevTableName : null;
+    String operation = Stream.of(database, operationType, schema, table, prevTable)
+                             .filter(s -> s != null && !s.isEmpty())
+                             .collect(Collectors.joining(", "));
+    return "{" + operation + "}";
```
sumitjnn: Similar to the above, we can use StringJoiner

sumitjnn (approved):

delta-api/src/main/java/io/cdap/delta/api/Offset.java
```diff
@@ -50,6 +50,13 @@ public boolean equals(Object o) {
     Offset offset = (Offset) o;
     return Objects.equals(state, offset.state);
   }
+  @Override
+  public String toString() {
+    if (state.isEmpty()) {
+      return null;
```
sumitjnn: We should not return null from toString as it can lead to unexpected NPE. Instead we can return string "state=null"

sumitjnn (changes requested):

delta-api/src/main/java/io/cdap/delta/api/Offset.java
yjhawar: I will remove this check. In case of null, it will return "{}" (empty map) while logging.Or can return an empty string.

delta-api/src/main/java/io/cdap/delta/api/DDLEvent.java
```diff
@@ -79,7 +79,7 @@ public String toString() {
     if (primaryKey != null && !primaryKey.isEmpty()) {
       stringJoiner.add("primaryKey=" + primaryKey);
     }
-    if (this.getOffset().toString() != null) {
+    if (!this.getOffset().toString().isEmpty()) {
```
sumitjnn: We should not write logic on return value of toString(), it's fine to just check if getOffset is not null

delta-api/src/main/java/io/cdap/delta/api/Offset.java
```diff
@@ -53,7 +53,7 @@ public boolean equals(Object o) {
   @Override
   public String toString() {
     if (state.isEmpty()) {
-      return null;
+      return "";
     }
```
sumitjnn: Let's remove this, it's fine to just call state.toString() even if state is an empty map

sumitjnn (changes requested):

sumitjnn (approved):

PR #220, opened 2023-03-24
Add toString() for DML event

yjhawar:


delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -121,6 +122,22 @@ public boolean equals(Object o) {
       Objects.equals(sortKeys, dmlEvent.sortKeys);
   }
 
+  @Override
+  public String toString() {
+    StringJoiner stringJoiner = new StringJoiner(", ", "{", "}");
+    stringJoiner.add("operation=" + operation);
+    stringJoiner.add("isSnapshot=" + this.isSnapshot());
+    stringJoiner.add("changeType=" + this.getChangeType());
+    stringJoiner.add("offset=" + this.getOffset());
```
sumitjnn, line 131: Is offset guaranteed to be not null ?

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
```diff
@@ -121,6 +122,22 @@ public boolean equals(Object o) {
...
+    if (this.getSourceTimestampMillis() != null) {
+      stringJoiner.add("sourceTimestampMillis=" + this.getSourceTimestampMillis());
+    }
+    if (sortKeys != null && !sortKeys.isEmpty()) {
```
sumitjnn, line 135: nit: we can use Collections.isEmpty instead

sumitjnn (approved):
Just few minor comments

delta-api/src/main/java/io/cdap/delta/api/DMLEvent.java
yjhawar: offset can be null in that case it returns an empty map.

delta-api/pom.xml
```diff
@@ -42,6 +42,12 @@
       <version>6.7.0-SNAPSHOT</version>
       <scope>compile</scope>
     </dependency>
+      <dependency>
+          <groupId>commons-collections</groupId>
+          <artifactId>commons-collections</artifactId>
```
sumitjnn: My bad, I thought we already had commons collection, let's not add a dependency just for one empty check

yjhawar: However CollectionUtils is a basic library which is generally used right ,so can always have basic use cases

sumitjnn: Yes it's a basic library but we should have good reason to add a dependency as it adds maintenance overhead like upgrades due to security vulnerabilities, I don't think replacing few null checks justifies it

sumitjnn (approved):

PR #221, opened 2023-03-28
Improving logging during exception handling (CDAP-20479)

yjhawar:
The following issues are handled :
 
- Fix any incorrect logging levels.
- Add original exception as cause where stacktrace isn't logged.
- Add necessary log messages if missing.

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/app/service/common/AbstractAssessorHandler.java
+++ b/delta-app/src/main/java/io/cdap/delta/app/service/common/AbstractAssessorHandler.java
@@ -64,6 +64,7 @@ public class AbstractAssessorHandler extends AbstractSystemHttpServiceHandler {
     try {
       NamespaceSummary namespaceSummary = context.getAdmin().getNamespaceSummary(namespaceName);
       if (namespaceSummary == null) {
+        LOG.error(String.format("Namespace '%s' not found", namespaceName));
         responder.sendError(HttpURLConnection.HTTP_NOT_FOUND, String.format("Namespace '%s' not found", namespaceName));
         return;
       }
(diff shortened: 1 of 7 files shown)


sumitjnn:
@yjhawar  Can you add jira id to PR title

delta-app/src/main/java/io/cdap/delta/app/service/common/AbstractAssessorHandler.java
```diff
@@ -64,6 +64,7 @@ protected void respond(String namespaceName, HttpServiceResponder responder, Nam
     try {
       NamespaceSummary namespaceSummary = context.getAdmin().getNamespaceSummary(namespaceName);
       if (namespaceSummary == null) {
+        LOG.error(String.format("Namespace '%s' not found", namespaceName));
```
sumitjnn: We should use the logger's build in parameter substitution
```suggestion
        LOG.error("Namespace {} not found", namespaceNam);
```

delta-app/src/main/java/io/cdap/delta/store/HCFSStateStore.java
```diff
@@ -42,6 +45,7 @@
  * Is used only during migration
  */
 public class HCFSStateStore implements StateStore {
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractAssessorHandler.class);
```
sumitjnn: Use correct class - HCFSStateStore

delta-app/src/main/java/io/cdap/delta/store/HCFSStateStore.java
```diff
@@ -120,6 +124,7 @@ public Collection<Integer> getWorkerInstances(DeltaPipelineId pipelineId) throws
           workerInstances.add(Integer.parseInt(fileName));
         } catch (NumberFormatException e) {
           // should not happen unless somebody manually modified the directories
+          LOG.error(String.format("Worker instance '%s' isn't valid.", fileName), e);
```
sumitjnn: use the logger's build in parameter substitution
`LOG.error("Worker instance {} is invalid.", fileName, e);`

sumitjnn (approved):

PR #228, opened 2023-05-04
CDAP-20613: Add capability to log thread dump for debugability

sumitjnn:
Thread dump is logged if
- Diagnostic mode is enabled via preference `diagnostic.mode.enabled` set to `true` (disabled by default)
- No events are processed by the target BQ plugin in a certain interval.

### DIFF
--- a/delta-app/src/main/java/io/cdap/delta/app/Diagnostics.java
+++ b/delta-app/src/main/java/io/cdap/delta/app/Diagnostics.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


delta-app/src/main/java/io/cdap/delta/app/Diagnostics.java
```diff
@@ -0,0 +1,48 @@
...
+    logThreadDump();
+  }
+
+  private static void logThreadDump() {
+    try {
+      ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
+      ThreadInfo[] threadInfos = threadMXBean.dumpAllThreads(true, true);
+
+      for (ThreadInfo threadInfo : threadInfos) {
+        LOG.info("" + threadInfo);
```
albertshau: does it need the `"" + ` for some reason?

albertshau (approved):

delta-app/src/main/java/io/cdap/delta/app/Diagnostics.java
sumitjnn: It's just to convert `threadInfo` to string as logger does not have any method which just accepts object. Guess we can just call `threadInfo.toString()` instead to avoid confusion.

PR #255, opened 2025-07-09
CDAP OSS migration - delta

sahusanket:


pom.xml
```diff
@@ -68,31 +68,6 @@
...
   </properties>
 
-  <repositories>
-    <repository>
-      <id>sonatype</id>
-      <url>https://oss.sonatype.org/content/groups/public</url>
-    </repository>
-  </repositories>
-
-  <pluginRepositories>
```
sahusanket, line 78: @itsankit-google we should remove these too ?

itsankit-google: we need to add snapshot repository instead, see here : https://github.com/data-integrations/google-cloud/blob/develop/pom.xml#L112

pom.xml
```diff
@@ -70,29 +70,30 @@
...
+      <id>sonatype-snapshots</id>
+      <url>https://central.sonatype.com/repository/maven-snapshots</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
     </pluginRepository>
   </pluginRepositories>
```
itsankit-google: we don't need `pluginRepositories` section anymore.

itsankit-google (approved):

